PR #9, opened 2021-02-20
Add serialization support for i/u128.

abizjak:


abizjak:
@limemloh This came up recently in discuss with one of the testers, and there is no real reason not to have it, although using u128 is a lot  more expensive than u64 so people should be encouraged to use the least type they can manage.

limemloh:
I think the schema should support this as well then.

abizjak:
> I think the schema should support this as well then.

I agree that we should at least have the SchemaType instance for it, although I am not sure whether we should add another type in the schema for it.

limemloh:
We could give it the SchemaType implementation of say a byte array.

But If we want the value to be displayed nicely by tools and allow the user to input as a JSON number, we need to extend the Schema.

gilescope:
This is cool - I tried using u128 in a contract but it wasn't supported so great to see this PR!

abizjak:
> We could give it the SchemaType implementation of say a byte array.
> 
> But If we want the value to be displayed nicely by tools and allow the user to input as a JSON number, we need to extend the Schema.

I added this now. It is not very satisfactory since it needs to be String in JSON due to limitations of the format/serde.

gilescope:
We are serialising to blockchain state as json? or is it really something like binary json? 

I seem to be going through a str/num phase at the moment so will take a look at serde's efficiency on that side.

abizjak:
> We are serialising to blockchain state as json? or is it really something like binary json?
> 
> I seem to be going through a str/num phase at the moment so will take a look at serde's efficiency on that side.

No, we are not serializing the state as JSON on the chain, i.e., smart contracts themselves are not doing it. But concordium-client and cargo-concordium can display the state as JSON if the schema is provided. That is the part I was referring to.

PR #10, opened 2021-02-22
bytes is a lot more efficient than chars.

gilescope:
I was recently improving to_digit in the std lib and came across this code and though we can do the same here. Unicode is complicated and here it isn't relevant.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -80,7 +80,7 @@ impl str::FromStr for Amount {
         let mut after_dot = 0;
         let mut state = 0;
         for c in v.bytes() {
-            let d = c - '0' as u8;
+            let d = c.wrapping_sub(b'0');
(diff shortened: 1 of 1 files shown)


gilescope:
Forgot - wrapping_sub is key.

abizjak:
When you say "a lot more efficient", what are we talking about?

This code is not used in any performance sensitive context, so clarity trumps slight efficiency differences.

To me it is clearer to talk about `char`s than bytes, the latter requiring us to deal with the details of encoding.

limemloh:
Thank you for spending time on this, but I will close this MR, because I think correctness and readability is more importing for this part of our code base, than the performance gain.

gilescope:
Sure - if it's not heavily used then that makes sense.

PR #14, opened 2021-03-09
CB-1203: Contract and receive names

Bargsteen:
Closes CB-1203.

Adds support for sending contract names and receive names as parameters.

Adds the schema types:
- `ContractName` 
- `ReceiveName`

Also adds the following data types:
- `ContractName`
- `OwnedContractName` (implements `SchemaType`)
- `ReceiveName`
- `OwnedReceiveName` (implements `SchemaType`)

The owned versions can be converted into their counterparts with `as_ref()`.

Related:
https://gitlab.com/Concordium/smart-contracts/-/merge_requests/125
https://gitlab.com/Concordium/concordium-client/-/merge_requests/291
https://github.com/Concordium/concordium-rust-smart-contracts/pull/25

### DIFF
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -723,13 +723,15 @@ mod impls {
                     Ok(Value::String(string))
                 }
                 Type::ContractName(size_len) => {
-                    let contract_name = OwnedContractName::new(deserial_string(source, size_len)?);
+                    let contract_name = OwnedContractName::new(deserial_string(source, size_len)?)
+                        .map_err(|_| ParseError::default())?;
                     let name_without_init =
                         contract_name.contract_name().ok_or_else(ParseError::default)?;
                     Ok(json!({ "contract": name_without_init }))
                 }
                 Type::ReceiveName(size_len) => {
-                    let receive_name = OwnedReceiveName::new(deserial_string(source, size_len)?);
+                    let receive_name = OwnedReceiveName::new(deserial_string(source, size_len)?)
+                        .map_err(|_| ParseError::default())?;
                     let contract_name =
                         receive_name.contract_name().ok_or_else(ParseError::default)?;
                     let func_name = receive_name.func_name().ok_or_else(ParseError::default)?;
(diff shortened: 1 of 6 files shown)


src/schema.rs
```diff
@@ -548,6 +574,35 @@ mod impls {
...
+                    bytes.extend_from_slice(&buf[..new]);
+                }
+            }
+            if read == len {
+                Ok(String::from_utf8(bytes)?)
+            } else {
+                Err(ParseError {})
+            }
+        }
+    }
```
abizjak, line 604: Isn't this duplicated from the Vec<u8> one? I think we should reuse that if possible and then use `String::from_utf8`

src/types.rs
```diff
@@ -598,6 +598,156 @@ pub enum Address {
...
+
+impl OwnedReceiveName {
+    /// Create a new OwnedReceiveName and check the format. Expected format:
+    /// "<contract_name>.<func_name>".
+    pub fn new(name: String) -> Result<Self, NewReceiveNameError> {
+        if !name.contains('.') {
+            return Err(NewReceiveNameError::MissingDotSeparator);
+        }
+        Ok(OwnedReceiveName(name))
+    }
```
abizjak, line 708: This should also check the length.

src/schema.rs
```diff
@@ -195,6 +197,14 @@ impl SchemaType for String {
     fn get_type() -> Type { Type::String(SizeLength::U32) }
 }
 
+impl SchemaType for OwnedContractName {
+    fn get_type() -> Type { Type::ContractName(SizeLength::U32) }
+}
+
+impl SchemaType for OwnedReceiveName {
+    fn get_type() -> Type { Type::ReceiveName(SizeLength::U32) }
+}
```
abizjak, line 206: Shouldn't we use U16 instead?

abizjak (changes requested):

src/schema.rs
abizjak: I had this strong sense of deja vu that I wrote this exact same code somewhere.

But it is just the factored out string indeed, not vec.

abizjak (dismissed):

abizjak (approved):

PR #15, opened 2021-03-12
Fix serialization of Contract- and ReceiveName

Bargsteen:
After switching to use SizeLength::U16 we can no longer use the
serialization from String.

### DIFF
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -326,7 +326,10 @@ impl Deserial for Address {
 
 impl<'a> Serial for ContractName<'a> {
     fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
-        self.get_chain_name().serial(out)
+        let name = self.get_chain_name();
+        let len = name.len() as u16;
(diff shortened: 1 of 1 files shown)


src/impls.rs
```diff
@@ -326,7 +326,10 @@ impl Deserial for Address {
 
 impl<'a> Serial for ContractName<'a> {
     fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
-        self.get_chain_name().serial(out)
+        let name = self.get_chain_name();
+        let len: u16 = name.len().try_into().map_err(|_| Default::default())?;
```
abizjak, line 332: I would just remove the `try_into` and use `as` since any value constructed should be shorter. ANd it is undefined what happens if you use `unchecked`.

abizjak (changes requested):
Just use `as`.

src/impls.rs
Bargsteen: Ok

PR #22, opened 2021-06-23
Adding better support for hashmaps

limemloh:
## Purpose

Support using `Address` in a `BTreeMap` and contract schemas.
Types implement Hash and add support for hashmaps in `no-std`.

## Changes

- Derive `PartialOrd` and `Ord` for `ContractAddress` and `Address`.
- Implement `SchemaType` for `Address`.
- Add `HashMap` and `HashSet` from the `hashbrown` crate to support `no-std`.
- Make `HashMap` and `HashSet` default to the `fnv` hasher.
- Derive `Hash` for `Amount`, `Timestamp`, `Duration`, `AccountAddress`, `ContractAddress`, `Address`, `ContractName`, `OwnedContractName`, `ReceiveName` and `OwnedReceiveName`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,11 +15,11 @@ readme = "README.md"
 arbitrary = { version = "0.4", features = ["derive"], optional = true }
 
 [dependencies.hashbrown]
-version = "0.11.2"
+version = "0.11"
 default-features = false
 
 [dependencies.fnv]
-version = "1.0.7"
+version = "1.0"
 default-features = false
 
 [dependencies.serde]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,14 @@ readme = "README.md"
 [dependencies]
 arbitrary = { version = "0.4", features = ["derive"], optional = true }
 
+[dependencies.hashbrown]
+version = "0.11.2"
+default-features = false
+
+[dependencies.fnv]
+version = "1.0.7"
+default-features = false
+
 [dependencies.serde]
 optional = true
 features = ["derive"]
@@ -34,7 +42,7 @@ version = "0.1"
 [features]
 default = ["std"]
 
-std = []
+std = ["fnv/std"]
 
 derive-serde = ["serde", "serde_json", "std", "base58check", "chrono"]
 
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -2,13 +2,12 @@ use crate::{constants::*, traits::*, types::*};
 
 #[cfg(not(feature = "std"))]
 use alloc::{boxed::Box, collections, string::String, vec::Vec};
+use collections::{BTreeMap, BTreeSet};
 #[cfg(not(feature = "std"))]
 use core::{hash, marker, mem::MaybeUninit, slice};
+use hash::Hash;
 #[cfg(feature = "std")]
 use std::{collections, hash, marker, mem::MaybeUninit, slice};
-
-use collections::{BTreeMap, BTreeSet};
-use hash::Hash;
 
 /// Apply the given macro to each of the elements in the list
 /// For example, `repeat_macro!(println, "foo", "bar")` is equivalent to
--- a/src/impls.rs
+++ b/src/impls.rs
@@ -411,7 +411,7 @@ pub fn deserial_vector_no_length<R: Read, T: Deserial>(
 
 /// Write a Map as a list of key-value pairs ordered by the key, without the
 /// length information.
-pub fn serial_map_no_length<'a, W: Write, K: Serial + 'a, V: Serial + 'a>(
+pub fn serial_map_no_length<W: Write, K: Serial, V: Serial>(
     map: &BTreeMap<K, V>,
     out: &mut W,
 ) -> Result<(), W::Err> {
@@ -471,7 +471,7 @@ pub fn deserial_map_no_length_no_order_check<R: Read, K: Deserial + Ord, V: Dese
 
 /// Write a HashMap as a list of key-value pairs in to particular order, without
 /// the length information.
-pub fn serial_hashmap_no_length<'a, W: Write, K: Serial + 'a, V: Serial + 'a>(
+pub fn serial_hashmap_no_length<W: Write, K: Serial, V: Serial>(
     map: &HashMap<K, V>,
     out: &mut W,
 ) -> Result<(), W::Err> {
@@ -499,7 +499,7 @@ pub fn deserial_hashmap_no_length<R: Read, K: Deserial + Eq + Hash, V: Deserial>
 }
 
 /// Write a [BTreeSet](https://doc.rust-lang.org/std/collections/struct.BTreeSet.html) as an ascending list of keys, without the length information.
-pub fn serial_set_no_length<'a, W: Write, K: Serial + 'a>(
+pub fn serial_set_no_length<W: Write, K: Serial>(
     map: &BTreeSet<K>,
     out: &mut W,
 ) -> Result<(), W::Err> {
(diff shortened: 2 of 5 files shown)


src/types.rs
```diff
@@ -2,9 +2,19 @@ use crate::constants;
 #[cfg(not(feature = "std"))]
 use alloc::{string::String, string::ToString, vec::Vec};
 #[cfg(not(feature = "std"))]
-use core::{convert, fmt, iter, ops, str};
+use core::{convert, fmt, hash, iter, ops, str};
 #[cfg(feature = "std")]
-use std::{convert, fmt, iter, ops, str};
+use std::{convert, fmt, hash, iter, ops, str};
+
```
abizjak, line 8: Empty line.

src/types.rs
```diff
@@ -600,14 +610,14 @@ pub struct ContractAddress {
     serde(tag = "type", content = "address", rename_all = "lowercase")
 )]
 #[cfg_attr(feature = "fuzz", derive(Arbitrary))]
-#[derive(PartialEq, Eq, Copy, Clone, Debug)]
+#[derive(PartialEq, Eq, Copy, Clone, PartialOrd, Ord, Debug, Hash)]
```
abizjak, line 613: Did you check what the generated comparison function is and whether we can make one that generates smaller code?

src/schema.rs
```diff
@@ -176,6 +178,14 @@ impl SchemaType for AccountAddress {
 impl SchemaType for ContractAddress {
     fn get_type() -> Type { Type::ContractAddress }
 }
+impl SchemaType for Address {
+    fn get_type() -> Type {
+        Type::Enum(Vec::from([
+            (String::from("account"), Fields::Unnamed(Vec::from([Type::AccountAddress]))),
+            (String::from("contract"), Fields::Unnamed(Vec::from([Type::ContractAddress]))),
```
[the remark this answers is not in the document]
limemloh: You mean the one where we have `{"type": "contract", "address": {...}}` for contracts and  `{"type": "account", "address": "...." }` for accounts? 
If so, then we would have to extend the contract schema to support this.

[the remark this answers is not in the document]
abizjak: Yes. It would look out of place as well, since we use a different method for encoding enums in JSON.

src/impls.rs
```diff
@@ -1,11 +1,14 @@
 use crate::{constants::*, traits::*, types::*};
 
 #[cfg(not(feature = "std"))]
-use alloc::{boxed::Box, collections::*, string::String, vec::Vec};
+use alloc::{boxed::Box, collections, string::String, vec::Vec};
```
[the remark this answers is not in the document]
limemloh: No, I don't think there are any changes to whatever is public from this.

Cargo.toml
```diff
@@ -14,6 +14,14 @@ readme = "README.md"
 [dependencies]
 arbitrary = { version = "0.4", features = ["derive"], optional = true }
 
+[dependencies.hashbrown]
+version = "0.11.2"
+default-features = false
+
+[dependencies.fnv]
+version = "1.0.7"
```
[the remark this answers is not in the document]
limemloh: I have removed the patch version

src/impls.rs
```diff
@@ -467,6 +469,35 @@ pub fn deserial_map_no_length_no_order_check<R: Read, K: Deserial + Ord, V: Dese
     Ok(out)
 }
 
+/// Write a HashMap as a list of key-value pairs in to particular order, without
+/// the length information.
+pub fn serial_hashmap_no_length<'a, W: Write, K: Serial + 'a, V: Serial + 'a>(
```
abizjak, line 474: What is this lifetime here for? I can't see it used for any of the arguments.

src/impls.rs
```diff
@@ -499,6 +530,33 @@ pub fn deserial_set_no_length<R: Read, K: Deserial + Ord + Copy>(
     Ok(out)
 }
 
+/// Write a [HashSet](https://doc.rust-lang.org/std/collections/struct.HashSet.html) as a list of keys in no particular order, without the length information.
+pub fn serial_hashset_no_length<'a, W: Write, K: Serial + 'a>(
```
abizjak, line 534: Same question about the lifetime.

src/impls.rs
```diff
@@ -565,6 +623,32 @@ impl<K: Deserial + Ord, V: Deserial> Deserial for BTreeMap<K, V> {
...
+/// They are serialized in no particular order.
+impl<K: Serial, V: Serial> Serial for HashMap<K, V> {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
+        let len = self.len() as u32;
+        len.serial(out)?;
+        serial_hashmap_no_length(self, out)
+    }
+}
+
+/// The deserialization of maps assumes their size as a u32.
```
abizjak, line 637:
```suggestion
/// The deserialization of maps assumes their size is a u32.
```

src/types.rs
```diff
@@ -600,14 +609,58 @@ pub struct ContractAddress {
...
+#[derive(Eq, Copy, Clone, Debug)]
 pub enum Address {
     Account(AccountAddress),
     Contract(ContractAddress),
 }
 
+// This trait is implemented manually to save bytes
+impl PartialEq for Address {
+    fn eq(&self, other: &Address) -> bool {
+        match (self, other) {
```
abizjak, line 621: Is there a reason for not pattern matching on `(&self, other)` here?

Is this better or does it not make a difference.

src/types.rs
```diff
@@ -600,14 +609,58 @@ pub struct ContractAddress {
...
 )]
 #[cfg_attr(feature = "fuzz", derive(Arbitrary))]
-#[derive(PartialEq, Eq, Copy, Clone, Debug)]
+#[derive(Eq, Copy, Clone, Debug)]
 pub enum Address {
     Account(AccountAddress),
     Contract(ContractAddress),
 }
 
+// This trait is implemented manually to save bytes
```
abizjak, line 618: Maybe rephrase this to something like 
"The trait is implemented manually so that the generated Wasm code is smaller."

src/types.rs
```diff
@@ -600,14 +609,58 @@ pub struct ContractAddress {
...
+    fn eq(&self, other: &Address) -> bool {
+        match (self, other) {
+            (Address::Account(s), Address::Account(o)) => s == o,
+            (Address::Contract(s), Address::Contract(o)) => s == o,
+            _ => false,
+        }
+    }
+}
+
+// This trait is implemented manually to save bytes
```
abizjak, line 629: And here.

src/types.rs
```diff
@@ -600,14 +609,58 @@ pub struct ContractAddress {
...
+            }
+            Address::Contract(address) => {
+                1u8.hash(state);
+                address.hash(state);
+            }
+        }
+    }
+}
+
+// This trait is implemented manually to save bytes
```
abizjak, line 645: And here.

src/types.rs
```diff
@@ -600,14 +609,58 @@ pub struct ContractAddress {
...
+impl PartialOrd for Address {
+    fn partial_cmp(&self, other: &Address) -> Option<Ordering> { Some(self.cmp(other)) }
+}
+
+// This trait is implemented manually to save bytes
+impl Ord for Address {
+    fn cmp(&self, other: &Address) -> Ordering {
+        match (self, other) {
```
abizjak, line 653: Same question about `(&self, other)`.

src/impls.rs
limemloh: I copied the signature from serial_map_no_length which also have this, but I'll make sure to remove these from both

src/types.rs
limemloh: I am not sure what you mean by this.
Would you mind elaborating?

abizjak: Never mind this, I read it wrong.

src/types.rs
abizjak: And the same here.

src/impls.rs
abizjak: It might have made sense at some point.

abizjak (dismissed):
If you would please fix the formatting, this is good to go.

abizjak (approved):

PR #33, opened 2021-12-07
Introduce types for parameters and entrypoints.

abizjak:
## Purpose

Introduce types for parameters and entrypoints.

## Changes

Introduce new types for entrypoints and parameters. The latter is useful to have a central place for its serialization.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -863,6 +863,15 @@ impl<'a> EntrypointName<'a> {
         is_valid_entrypoint_name(name)?;
         Ok(Self(name))
     }
+
+    /// Create a new name. **This does not check the format and is therefore
+    /// unsafe.** It is provided for convenience since sometimes it is
+    /// statically clear that the format is satisfied.
+    pub fn new_unchecked(name: &'a str) -> Self { Self(name) }
+}
+
+impl<'a> From<EntrypointName<'a>> for &'a str {
+    fn from(en: EntrypointName<'a>) -> Self { en.0 }
 }
 
 impl<'a> From<EntrypointName<'a>> for OwnedEntrypointName {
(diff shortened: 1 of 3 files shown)


src/types.rs
```diff
@@ -848,6 +848,71 @@ impl OwnedReceiveName {
...
+}
+
+impl<'a> From<EntrypointName<'a>> for OwnedEntrypointName {
+    fn from(epn: EntrypointName<'a>) -> Self { Self(String::from(epn.0)) }
+}
+
+/// An entrypoint name (owned version). Expected format:
+/// "<func_name>" where
+#[derive(Eq, PartialEq, Debug, Clone, Hash)]
+pub struct OwnedEntrypointName(pub String);
```
Bargsteen, line 875: I think this should be `pub(crate)` so you are forced to use the `new` method.

```suggestion
pub struct OwnedEntrypointName(pub(crate) String);
```

src/types.rs
```diff
@@ -848,6 +848,71 @@ impl OwnedReceiveName {
...
 }
 
+/// An entrypoint name (borrowed version). Expected format:
+/// "<func_name>" where
+#[derive(Eq, PartialEq, Debug, Clone, Copy, Hash)]
+pub struct EntrypointName<'a>(pub(crate) &'a str);
+
+impl<'a> EntrypointName<'a> {
+    /// Size of the name in bytes.
+    pub fn size(&self) -> u32 { self.0.as_bytes().len() as u32 }
```
Bargsteen, line 858: This should be equivalent. Did you add the `as_bytes()` to be more explicit?
```suggestion
    pub fn size(&self) -> u32 { self.0.len() as u32 }
```

Bargsteen (changes requested):
Looks good apart from a few corrections and comments.

src/types.rs
abizjak: Yes, indeed (both for they are equivalent, and I added it to be explicit). length of strings is always ambiguous, so I thought it was better to be explicit.

src/types.rs
abizjak: I tend to agree, but I'll double check whether this was not needed in some ffi (otherwise we might need to add some other From/To instance or similar).

src/types.rs
Bargsteen: Right, then I think we should do the same in
```
    if name.len() > constants::MAX_FUNC_NAME_SIZE {
```

src/types.rs
abizjak: Changed, and From implementations added.

Bargsteen (approved):
Looks good! 👍

PR #34, opened 2022-01-24
Add schema for V1 contracts

abizjak:
## Purpose

Add support for V1 schemas.

## Changes

Add a return value schema for V1 contracts.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -73,6 +73,32 @@ pub enum FunctionSchema {
     },
 }
 
+impl FunctionSchema {
+    /// Extract the parameter schema if it exists.
+    pub fn parameter(&self) -> Option<&Type> {
+        match self {
+            FunctionSchema::Parameter(ty) => Some(ty),
+            FunctionSchema::ReturnValue(_) => None,
+            FunctionSchema::Both {
+                parameter,
+                ..
+            } => Some(parameter),
+        }
+    }
+
+    /// Extract the return value schema if it exists.
+    pub fn return_value(&self) -> Option<&Type> {
+        match self {
+            FunctionSchema::Parameter(_) => None,
+            FunctionSchema::ReturnValue(rv) => Some(rv),
+            FunctionSchema::Both {
+                return_value,
+                ..
+            } => Some(return_value),
+        }
+    }
+}
+
 /// Schema for the fields of a struct or some enum variant.
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(test, derive(arbitrary::Arbitrary))]
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -33,20 +33,36 @@ pub trait SchemaType {
     fn get_type() -> crate::schema::Type;
 }
 
-/// Contains all the contract schemas for a module
+/// Contains all the contract schemas for a V0 module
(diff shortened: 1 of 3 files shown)


src/schema.rs
```diff
@@ -272,24 +314,41 @@ impl Deserial for Fields {
...
         self.contracts.serial(out)?;
         Ok(())
     }
 }
 
-impl Deserial for Module {
+impl Deserial for ModuleV0 {
     fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
         let len: u32 = source.get()?;
         let contracts = deserial_map_no_length_no_order_check(source, len as usize)?;
```
Bargsteen, line 334: I think we should just use `deserial`, which does the exact same thing.

src/schema.rs
```diff
@@ -272,24 +314,41 @@ impl Deserial for Fields {
...
             contracts,
         })
     }
 }
 
-impl Serial for Contract {
+impl Deserial for ModuleV1 {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        let len: u32 = source.get()?;
+        let contracts = deserial_map_no_length_no_order_check(source, len as usize)?;
```
Bargsteen, line 344: And here as well.

src/schema.rs
```diff
@@ -33,26 +33,68 @@ pub trait SchemaType {
...
+#[derive(Debug, Default, Clone)]
+/// The [Default] instance produces an empty schema.
+pub struct ContractV1 {
+    pub init:    Option<FunctionSchema>,
+    pub receive: BTreeMap<String, FunctionSchema>,
+}
+
+/// Describes the schema of an init or a receive function for V1 contracts.
+#[derive(Debug, Clone)]
+pub enum FunctionSchema {
```
Bargsteen, line 67: I think this should be named `Function` to be consistent with the rest. We usually use it qualified anyhow, so `schema::Function` is clear on its own.

src/schema.rs
```diff
@@ -298,20 +357,78 @@ impl Serial for Contract {
...
+        Ok(())
+    }
+}
+
+impl Deserial for ContractV0 {
     fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
         let state = source.get()?;
         let init = source.get()?;
         let len: u32 = source.get()?;
         let receive = deserial_map_no_length_no_order_check(source, len as usize)?;
```
Bargsteen, line 373: And here

src/schema.rs
```diff
@@ -298,20 +357,78 @@ impl Serial for Contract {
...
             receive,
         })
     }
 }
 
+impl Deserial for ContractV1 {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        let init = source.get()?;
+        let len: u32 = source.get()?;
+        let receive = deserial_map_no_length_no_order_check(source, len as usize)?;
```
Bargsteen, line 386: And here

Bargsteen (approved):
LGTM, just minor tweaks needed.

src/schema.rs
abizjak: I don't particularly want to change this. I think it is OK to be explicit here.

PR #35, opened 2022-02-16
Auto-serializing constructor for `OwnedParameter`

Bargsteen:
## Purpose

Adds an auto-serializing constructor for `OwnedParameter` similar to what we had with the `send` helper function.

## Changes

Add two constructors to `OwnedParameter`

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -934,6 +934,7 @@ impl<'a> AsRef<[u8]> for OwnedParameter {
     fn as_ref(&self) -> &[u8] { self.0.as_ref() }
 }
 
+/// Convert the vector into a parameter as is.
 impl From<Vec<u8>> for OwnedParameter {
     #[inline(always)]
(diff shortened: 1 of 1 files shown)


src/types.rs
```diff
@@ -941,6 +941,13 @@ impl From<Vec<u8>> for OwnedParameter {
 
 impl OwnedParameter {
     pub fn as_parameter(&self) -> Parameter { Parameter(self.0.as_ref()) }
+
+    /// Construct an `OwnedParameter` by serializing the input using its
+    /// `Serial` instance.
+    pub fn new<D: Serial>(data: &D) -> Self { Self(to_bytes(data)) }
+
+    /// Construct an `OwnedParameter` from the raw, serialized, data.
+    pub fn new_raw(data: Vec<u8>) -> Self { Self(data) }
```
abizjak, line 950: There is already a `From` implementation that does exactly this so I would rather remove this.

PR #36, opened 2022-02-24
Add a few helper instances related to receive/entrypoint names.

abizjak:
## Purpose

Minor helpers and instances that are needed in cargo-concordium.

## Changes

The main change is moving some methods from OwnedReceiveName to ReceiveName since they are sometimes needed there, and since it is cheap to convert to the latter from the former, 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -3,26 +3,29 @@ use crate::types::*;
 use alloc::vec::Vec;
 use core::{default::Default, mem::MaybeUninit, slice};
 
-/// This is the equivalent to the
+/// This is essentially equivalent to the
 /// [SeekFrom](https://doc.rust-lang.org/std/io/enum.SeekFrom.html) type from
 /// the rust standard library, but reproduced here to avoid dependency on
-/// `std::io`.
+/// `std::io`, as well as to use 32-bit integers to specify positions. This
+/// saves some computation and space, and is adequate for the kind of data sizes
+/// that are possible in smart contracts.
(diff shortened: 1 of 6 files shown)


Bargsteen (dismissed):
Looks good in general.

However, I think we should make the methods easier to find on `OwnedReceiveName`.

Ideally, we could implement `Deref<Target=ReceiveName<'a>> for OwnedReceiveName`, but I don't think that's so easy due to the lifetime.

A decent alternative is to add methods like `func_name() { self.as_receive_name().func_name() }` to `OwnedReceiveName`.

src/traits.rs
```diff
@@ -3,26 +3,29 @@ use crate::types::*;
...
+    End(i32),
+    Current(i32),
 }
 
 /// The `Seek` trait provides a cursor which can be moved within a stream of
 /// bytes. This is essentially a copy of
 /// [std::io::Seek](https://doc.rust-lang.org/std/io/trait.Seek.html), but
 /// avoiding its dependency on `std::io::Error`, and the associated code size
-/// increase.
+/// increase. Additionally, the positions are expressed in terms on 32-bit
```
Bargsteen, line 22:
```suggestion
/// increase. Additionally, the positions are expressed in terms of 32-bit
```

Bargsteen (dismissed):

src/traits.rs
abizjak: Indeed.

PR #37, opened 2022-03-03
Quickcheck

simonkamp:
## Purpose

Adds support for using the QuickCheck crate to do property based testing.

## Changes

This adds implementations of the `quickcheck::Arbitrary` trait for smart contract types, including defining instances from random bytes and shrinking of instances.
The added code compiles when the `concordium-quickcheck` feature is enabled.

## Checklist

- [X] I have performed a self-review of the changes.
      hard-to-understand areas.


CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 - Introduce Entrypoint and Parameter types, and their owned versions.
 - Add a new schema version for V1 smart contracts. 
   This adds schema for return values of init and receive functions, and removes the state schema.
+- Implement `quickcheck::arbitrary` for `Timestamp`, `AccountAddress`, `ContractAddress`, `Address`,
+  `ChainMetadata`, `AttributeTag`, and `OwnedPolicy`.
```
abizjak, line 9: This should go under the Unreleased changes section.

Cargo.toml
```diff
@@ -42,7 +44,7 @@ version = "0.1"
 [features]
 default = ["std"]
 
-std = ["fnv/std"]
+std = ["fnv/std", "getrandom", "quickcheck"]
```
abizjak, line 47: I think we should have a separate feature for this. We don't want std to always enable this.

That feature can depend on std of course, if it is necessary.

src/types.rs
```diff
@@ -32,6 +34,38 @@ pub struct Amount {
...
+            micro_ccd: quickcheck::Arbitrary::arbitrary(g),
+        }
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        Box::new(quickcheck::Arbitrary::shrink(&self.micro_ccd).map(|mccd| Amount {
+            micro_ccd: mccd,
+        }))
+    }
+}
```
abizjak, line 67: This looks like something we could derive automatically, does it not? https://docs.rs/arbitrary/1.1.0/arbitrary/trait.Arbitrary.html#deriving-arbitrary

Is there a reason I'm not seeing for doing it manually?

src/types.rs
```diff
@@ -585,6 +634,18 @@ impl fmt::Display for Duration {
 #[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub struct AccountAddress(pub [u8; ACCOUNT_ADDRESS_SIZE]);
 
+#[cfg(feature = "std")]
+impl quickcheck::Arbitrary for AccountAddress {
+    fn arbitrary(g: &mut Gen) -> AccountAddress {
+        //todo should it satisfy any properties, or are random bytes okay?
```
abizjak, line 640: Random bytes are indeed OK. There are no requirements for the address in this format. 

But again, could we not just derive this?

src/types.rs
```diff
@@ -1073,6 +1189,16 @@ pub struct Cursor<T> {
 #[derive(Clone, Copy, Debug, Ord, PartialOrd, Eq, PartialEq)]
 pub struct AttributeTag(pub u8);
 
+#[cfg(feature = "std")]
+impl quickcheck::Arbitrary for AttributeTag {
+    //todo it is not clear to me if random instances of these are useful in tests
+    fn arbitrary(g: &mut Gen) -> AttributeTag { AttributeTag(quickcheck::Arbitrary::arbitrary(g)) }
```
abizjak, line 1195: I think you are right. We should only generate valid attributetags, valid in the sense that they can appear in valid attribute lists. That should only be 0-14 I think (the definitive list is in the rust-src/id/src/types.rs in concordium-base).

src/types.rs
```diff
@@ -1120,6 +1246,36 @@ pub struct Policy<Attributes> {
...
 }
 
+#[cfg(feature = "std")]
+impl quickcheck::Arbitrary for OwnedPolicy {
+    //todo it is not clear to me if random instances of these are useful in tests
+    fn arbitrary(g: &mut Gen) -> OwnedPolicy {
+        OwnedPolicy {
+            identity_provider: quickcheck::Arbitrary::arbitrary(g),
+            created_at:        quickcheck::Arbitrary::arbitrary(g),
+            valid_to:          quickcheck::Arbitrary::arbitrary(g),
```
abizjak, line 1256: It would be useful if valid_to >= created_at always.

src/types.rs
```diff
@@ -1059,6 +1160,21 @@ pub struct ChainMetadata {
...
+            slot_time: quickcheck::Arbitrary::arbitrary(g),
+        }
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        Box::new(quickcheck::Arbitrary::shrink(&self.slot_time).map(|st| ChainMetadata {
+            slot_time: st,
+        }))
+    }
+}
```
abizjak, line 1176: Couldn't we derive this?

src/types.rs
```diff
@@ -1120,6 +1246,36 @@ pub struct Policy<Attributes> {
...
 
+#[cfg(feature = "std")]
+impl quickcheck::Arbitrary for OwnedPolicy {
+    //todo it is not clear to me if random instances of these are useful in tests
+    fn arbitrary(g: &mut Gen) -> OwnedPolicy {
+        OwnedPolicy {
+            identity_provider: quickcheck::Arbitrary::arbitrary(g),
+            created_at:        quickcheck::Arbitrary::arbitrary(g),
+            valid_to:          quickcheck::Arbitrary::arbitrary(g),
+            items:             quickcheck::Arbitrary::arbitrary(g),
```
abizjak, line 1257: We should not generate arbitrary vectors here. Attribute values can only have limited length. If I recall correctly it's 31 bytes (the definitive value is in rust-src/id/src/constants.rs in concordium-base).

src/types.rs
```diff
@@ -32,6 +34,38 @@ pub struct Amount {
     pub micro_ccd: u64,
 }
 
+#[cfg(all(feature = "std", target_arch = "wasm32"))]
+use std::sync::atomic::AtomicU8;
+#[cfg(all(feature = "std", target_arch = "wasm32"))]
+static ADDRESS_COUNTER: AtomicU8 = AtomicU8::new(0);
```
abizjak, line 40: Have you tested that this works with our cargo-concordium?

src/types.rs
```diff
@@ -32,6 +34,38 @@ pub struct Amount {
...
+#[cfg(all(feature = "std", target_arch = "wasm32"))]
+fn custom_random(dest: &mut [u8]) -> Result<(), getrandom::Error> {
+    for byte in dest.iter_mut() {
+        *byte = ADDRESS_COUNTER.load(std::sync::atomic::Ordering::SeqCst);
+        ADDRESS_COUNTER.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+    }
+    Ok(())
+}
+#[cfg(all(feature = "std", target_arch = "wasm32"))]
+register_custom_getrandom!(custom_random);
```
abizjak, line 52: Could we instead make use of special host functions for this?

Host functions that are only available during testing with cargo-concordium-test. The same way that we currently do with `report_error`. The idea being that `cargo-concordium` selects the seed and then generates random values from it and we register that host function with quickcheck here.

Would that work? We can discuss the details if my message above is cryptic.

Cargo.toml
annenkov: Done

CHANGELOG.md
annenkov: Done

src/types.rs
annenkov: Turns out [arbitrary::Arbitrary](https://docs.rs/arbitrary/1.1.0/arbitrary/trait.Arbitrary.html#deriving-arbitrary) is not the same as [quickcheck::Arbitrary](https://docs.rs/quickcheck/latest/quickcheck/trait.Arbitrary.html). The one from quickcheck doesn't have built-in `derive`. [This crate](https://github.com/panicbit/quickcheck_derive.) looks outdated and I couldn't make it work.

Also, the PR adding this feature was [abandoned](https://github.com/BurntSushi/quickcheck/pull/222#issuecomment-573267560). So, I guess we just keep these instances.

src/types.rs
annenkov: Done

src/types.rs
annenkov: Done

src/types.rs
annenkov: Done

concordium-contracts-common/src/types.rs
```diff
@@ -41,6 +45,21 @@ pub struct Amount {
...
+    fn arbitrary(g: &mut Gen) -> Amount {
+        Amount {
+            micro_ccd: quickcheck::Arbitrary::arbitrary(g),
+        }
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        Box::new(quickcheck::Arbitrary::shrink(&self.micro_ccd).map(|mccd| Amount {
+            micro_ccd: mccd,
+        }))
```
Bargsteen, line 59: This should also work, and I think it's nicer
```suggestion
        Box::new(quickcheck::Arbitrary::shrink(&self.micro_ccd).map(Amount::from_micro_ccd))
```

concordium-contracts-common/src/types.rs
```diff
@@ -362,6 +381,21 @@ pub struct Timestamp {
...
+#[cfg(feature = "concordium-quickcheck")]
+impl quickcheck::Arbitrary for Timestamp {
+    fn arbitrary(g: &mut Gen) -> Timestamp {
+        Timestamp {
+            milliseconds: quickcheck::Arbitrary::arbitrary(g),
+        }
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        Box::new(quickcheck::Arbitrary::shrink(&self.milliseconds).map(|ms| Timestamp {
```
Bargsteen, line 393: Same as above, but with `Timestamp::from_timestamp_millis`

concordium-contracts-common/src/types.rs
```diff
@@ -627,6 +661,18 @@ impl fmt::Display for Duration {
 #[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub struct AccountAddress(pub [u8; ACCOUNT_ADDRESS_SIZE]);
 
+#[cfg(feature = "concordium-quickcheck")]
+impl quickcheck::Arbitrary for AccountAddress {
+    fn arbitrary(g: &mut Gen) -> AccountAddress {
+        //todo should it satisfy any properties, or are random bytes okay?
```
Bargsteen, line 667: TODO.

To my knowledge, random bytes should be ok. We don't check the version or checksum in contracts.
On-chain, the first byte is the version (always `1` currently), the next 27 bytes are the payload, and the last four bytes are the first four bytes of `sha256(sha256(version + payload))`. More info here: https://en.bitcoin.it/wiki/Base58Check_encoding (bitcoin uses `0` for the version).

concordium-contracts-common/src/types.rs
```diff
@@ -1289,6 +1405,20 @@ impl AttributeValue {
...
+#[cfg(feature = "concordium-quickcheck")]
+impl quickcheck::Arbitrary for AttributeValue {
+    fn arbitrary(g: &mut Gen) -> AttributeValue {
+        AttributeValue {
+            inner: [0u8; 32].map(|_| quickcheck::Arbitrary::arbitrary(g)),
+        }
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        // todo: implement a shinker if it makes sense
```
Bargsteen, line 1417: The first byte encodes the length of the attribute, so you shrink the length and either set the remaining bytes to 0 or leave it as is.

concordium-contracts-common/src/types.rs
```diff
@@ -1406,6 +1536,67 @@ pub struct Policy<Attributes> {
...
+const ATTRIBUTE_TAG_VEC_LENGTH: [u8; 15] = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14];
+#[cfg(feature = "concordium-quickcheck")]
+fn gen_range_u64(g: &mut Gen, range: core::ops::Range<u64>) -> u64 {
+    let i: u64 = quickcheck::Arbitrary::arbitrary(g);
+    i % (range.end - range.start) + range.start
+}
+
+#[cfg(feature = "concordium-quickcheck")]
+impl quickcheck::Arbitrary for OwnedPolicy {
+    //todo it is not clear to me if random instances of these are useful in tests
```
Bargsteen, line 1564: TODO.
Not sure either, perhaps Ales has some input.

Bargsteen (changes requested):
Nice work, just missing a few minor things as I see it 😊

concordium-contracts-common/src/types.rs
annenkov: I think it's an old comment. The current implementation should generate random, but valid instances.

concordium-contracts-common/src/types.rs
```diff
@@ -41,6 +45,21 @@ pub struct Amount {
     pub micro_ccd: u64,
 }
 
+#[cfg(feature = "concordium-quickcheck")]
+impl quickcheck::Arbitrary for Amount {
+    fn arbitrary(g: &mut Gen) -> Amount {
+        Amount {
+            micro_ccd: quickcheck::Arbitrary::arbitrary(g),
+        }
```
abizjak, line 53:
```suggestion
        Amount::from_micro_ccd(quickcheck::Arbitrary::arbitrary(g))
```

concordium-contracts-common/src/types.rs
```diff
@@ -362,6 +381,21 @@ pub struct Timestamp {
     pub(crate) milliseconds: u64,
 }
 
+#[cfg(feature = "concordium-quickcheck")]
+impl quickcheck::Arbitrary for Timestamp {
+    fn arbitrary(g: &mut Gen) -> Timestamp {
+        Timestamp {
+            milliseconds: quickcheck::Arbitrary::arbitrary(g),
+        }
```
abizjak, line 389: As Kasper suggests, please use `from_timestamp_millis` instead.

concordium-contracts-common/src/types.rs
abizjak: The account address is by definition a SHA256(credId) where credId is the credential registration ID of the first credential (it is 48 bytes, and an encoding of some group element).

Hence an account address has no structure.

The String representation of the account address uses base58check encoding, and that one is defined by prepending a 1 byte, and hashing the result twice, then appending the last 4 bytes of the hash. Then encoding that in base58.

But an account address itself has no structure.

concordium-contracts-common/src/types.rs
```diff
@@ -627,6 +661,18 @@ impl fmt::Display for Duration {
...
 pub struct AccountAddress(pub [u8; ACCOUNT_ADDRESS_SIZE]);
 
+#[cfg(feature = "concordium-quickcheck")]
+impl quickcheck::Arbitrary for AccountAddress {
+    fn arbitrary(g: &mut Gen) -> AccountAddress {
+        //todo should it satisfy any properties, or are random bytes okay?
+        let mut bytes = [0u8; ACCOUNT_ADDRESS_SIZE];
+        for byte in bytes.iter_mut() {
+            *byte = quickcheck::Arbitrary::arbitrary(g);
+        }
```
abizjak, line 671: Doesn't quickcheck have an arbitrary instance for arrays?

concordium-contracts-common/src/types.rs
```diff
@@ -677,6 +723,25 @@ impl ContractAddress {
...
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        let index_shrink = quickcheck::Arbitrary::shrink(&self.index);
+        let subindex_shrink = quickcheck::Arbitrary::shrink(&self.subindex);
+        Box::new(index_shrink.zip(subindex_shrink).map(|(i, si)| ContractAddress {
+            index:    i,
+            subindex: si,
+        }))
```
abizjak, line 741: What is the idea here?

Don't we want something like a product of those two iterators? Why is zipping them the sensible thing?

concordium-contracts-common/src/types.rs
```diff
@@ -1250,6 +1351,21 @@ impl fmt::Display for NewAttributeValueError {
 #[derive(Clone, Copy, Debug, Ord, PartialOrd, Eq, PartialEq)]
 pub struct AttributeTag(pub u8);
 
+#[cfg(feature = "concordium-quickcheck")]
+impl quickcheck::Arbitrary for AttributeTag {
+    // We generate only valid tags, that is indices into the `ATTRIBUTE_NAMES` array
+    // defined in `concordium-base/rust-src/id/src/types.rs`
```
abizjak, line 1357: Why is that the choice that is made?

We discussed that it might be sensible to generate arbitrary tags to make the contract more robust with respect to future additions.

concordium-contracts-common/src/types.rs
```diff
@@ -1406,6 +1536,67 @@ pub struct Policy<Attributes> {
...
+    size: usize,
+) -> Vec<(A, B)> {
+    use std::collections::BTreeMap;
+    let mut m: BTreeMap<A, B> = BTreeMap::new();
+    while m.len() < size {
+        let k = A::arbitrary(g);
+        let v = B::arbitrary(g);
+        m.insert(k, v);
+    }
+    m.into_iter().collect()
```
abizjak, line 1551: Perhaps it would make sense to relax this to avoid infinite loops. And just generate vectors of size <= m.

```
for _ in 0..m {

        let k = A::arbitrary(g);
        let v = B::arbitrary(g);
        m.insert(k, v);
}
```

concordium-contracts-common/src/types.rs
abizjak: We should remove the todo. The instances are useful.

concordium-contracts-common/src/types.rs
annenkov: Done

concordium-contracts-common/src/types.rs
annenkov: Done

concordium-contracts-common/src/types.rs
annenkov: So, it's ok to keep random bytes, right?

abizjak: Yes.

concordium-contracts-common/src/types.rs
annenkov: @Bargsteen If the first byte is the length, should then `arbitrary`  take this into account? Could you elaborate on how it works exactly? Does the first byte determine how many remaining bytes (out of 32) to consider?

Bargsteen: I mean, that is up to us, but that was what I suggested. Yes, the length describes how many of the remaining 31 bytes contain actual data.

abizjak: The encoding is pretty precise. Attribute values are never longer than 31 bytes.

And it matters for the meaning what the length is. So please generate
- first byte is a value between 0 and 31, inclusive
- followed by this many bytes of arbitrary data.

concordium-contracts-common/src/types.rs
annenkov: Done. I left it as it is and removed the comment.

concordium-contracts-common/src/types.rs
annenkov: I don't think so, I looked at the source code, but couldn't find any.

concordium-contracts-common/src/types.rs
annenkov: This comes from Simon's code. I agree, it should be a product. Is there an std function that creates a product of iterators? I could only find an external library https://docs.rs/itertools/0.10.5/itertools/structs/struct.Product.html

annenkov: Yes, fair enough, I'll use nested loops. Just thought that maybe something standard is available, so I don't reinvent the wheel :)

concordium-contracts-common/src/types.rs
annenkov: Done.

concordium-contracts-common/src/types.rs
```diff
@@ -1406,6 +1536,67 @@ pub struct Policy<Attributes> {
     pub items:             Attributes,
 }
 
+#[cfg(feature = "concordium-quickcheck")]
+fn gen_no_dup_kv_vec<A: quickcheck::Arbitrary + Ord, B: quickcheck::Arbitrary>(
```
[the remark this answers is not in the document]
annenkov: Fixed as suggested by @abizjak and added a comment.

concordium-contracts-common/src/types.rs
annenkov: Done here an for `OwnedPolicy` as well.

concordium-contracts-common/src/types.rs
annenkov: Done.

concordium-contracts-common/src/types.rs
annenkov: Yes, indeed. I relaxed the constraint, so a tag is just a random `u8` now.

concordium-contracts-common/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+- Implement `quickcheck::Arbitrary` for `Timestamp`, `AccountAddress`, `ContractAddress`, `Address`,  `ChainMetadata`, `AttributeTag`, `AttributeValue` and `OwnedPolicy`.
```
abizjak, line 5: Please merge **main** into this branch and resolve the conflict here. The last contracts-common release is 5.0.0

concordium-contracts-common/src/types.rs
```diff
@@ -692,6 +750,27 @@ pub enum Address {
...
+        }
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        match self {
+            Address::Account(a) => Box::new(quickcheck::Arbitrary::shrink(a).map(Address::Account)),
+            Address::Contract(a) => {
+                Box::new(quickcheck::Arbitrary::shrink(a).map(Address::Contract))
+            }
+        }
```
abizjak, line 770: Please explain that we make a choice here that when shrinking an account address we only produce account addresses, and vice-versa.

This is a reasonable choice, but not the only reasonable choice :)

concordium-contracts-common/src/types.rs
```diff
@@ -1215,6 +1294,21 @@ pub struct ChainMetadata {
...
+    fn arbitrary(g: &mut Gen) -> ChainMetadata {
+        ChainMetadata {
+            slot_time: quickcheck::Arbitrary::arbitrary(g),
+        }
+    }
+
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
+        Box::new(quickcheck::Arbitrary::shrink(&self.slot_time).map(|st| ChainMetadata {
+            slot_time: st,
+        }))
```
abizjak, line 1308:
```suggestion
        Box::new(quickcheck::Arbitrary::shrink(&self.slot_time).map(|slot_time| ChainMetadata {
            slot_time,
        }))
```

concordium-contracts-common/src/types.rs
```diff
@@ -1250,6 +1344,15 @@ impl fmt::Display for NewAttributeValueError {
 #[derive(Clone, Copy, Debug, Ord, PartialOrd, Eq, PartialEq)]
 pub struct AttributeTag(pub u8);
 
+#[cfg(feature = "concordium-quickcheck")]
+impl quickcheck::Arbitrary for AttributeTag {
+    fn arbitrary(g: &mut Gen) -> AttributeTag { AttributeTag(quickcheck::Arbitrary::arbitrary(g)) }
```
abizjak, line 1349: Please document the discussions we had here about why we do not only generate tags 0-14.

concordium-contracts-common/src/types.rs
```diff
@@ -1406,6 +1545,97 @@ pub struct Policy<Attributes> {
...
+        } => created_at <= valid_to,
+    }
+}
+
+#[cfg(feature = "concordium-quickcheck")]
+impl quickcheck::Arbitrary for OwnedPolicy {
+    fn arbitrary(g: &mut Gen) -> OwnedPolicy {
+        let size: u8 = quickcheck::Arbitrary::arbitrary(g);
+        let created_at: Timestamp = quickcheck::Arbitrary::arbitrary(g);
+        // generate `valid_to` date so it's <= `created_at`
```
abizjak, line 1600: Probably the other way around :)

concordium-contracts-common/src/types.rs
```diff
@@ -1406,6 +1545,97 @@ pub struct Policy<Attributes> {
...
+                            valid_to:          vt,
+                            items:             it,
+                        })
+                    })
+                })
+            })
+            .flatten()
+            .flatten()
+            .flatten()
+            .filter(valid_owned_policy);
```
abizjak, line 1634: Why can't we use `flat_map` instead of `map + flatten`?

abizjak (commented):
Looks good overall. Just a few minor issues remaining.

concordium-contracts-common/CHANGELOG.md
annenkov: Done

concordium-contracts-common/src/types.rs
annenkov: Done.

concordium-contracts-common/src/types.rs
annenkov: Done.

concordium-contracts-common/src/types.rs
annenkov: Fixed.

concordium-contracts-common/src/types.rs
annenkov: Done.

concordium-contracts-common/src/types.rs
```diff
@@ -1250,6 +1346,20 @@ impl fmt::Display for NewAttributeValueError {
 #[derive(Clone, Copy, Debug, Ord, PartialOrd, Eq, PartialEq)]
 pub struct AttributeTag(pub u8);
 
+#[cfg(feature = "concordium-quickcheck")]
+impl quickcheck::Arbitrary for AttributeTag {
+    // We choose not to consttain the generated attributes to currently defined in
```
abizjak, line 1351:
```suggestion
    // We choose not to constrain the generated attributes to currently defined in
```

abizjak (approved):

concordium-contracts-common/src/types.rs
```diff
@@ -1250,6 +1346,20 @@ impl fmt::Display for NewAttributeValueError {
 #[derive(Clone, Copy, Debug, Ord, PartialOrd, Eq, PartialEq)]
 pub struct AttributeTag(pub u8);
 
+#[cfg(feature = "concordium-quickcheck")]
+impl quickcheck::Arbitrary for AttributeTag {
+    // We choose not to constrain the generated attributes to currently defined in
```
Bargsteen, line 1351:
```suggestion
    // We choose not to constrain the generated attributes to those currently defined in
```

concordium-contracts-common/src/types.rs
```diff
@@ -1406,6 +1552,94 @@ pub struct Policy<Attributes> {
...
+/// value using the `%` operator. The reason for doing it this way is that the
+/// range generation method is not exposed by QuickCheck's `Gen`.
+#[cfg(feature = "concordium-quickcheck")]
+fn gen_range_u8(g: &mut Gen, range: core::ops::Range<u8>) -> u8 {
+    let i: u8 = quickcheck::Arbitrary::arbitrary(g);
+    i % (range.end - range.start) + range.start
+}
+
+#[cfg(feature = "concordium-quickcheck")]
+fn valid_owned_policy(op: &OwnedPolicy) -> bool {
```
Bargsteen, line 1591: Please document this function with a one liner.

Bargsteen (approved):
Nice work 😊

concordium-contracts-common/src/types.rs
annenkov: Done.

PR #42, opened 2022-06-08
Add Leb128 integers and bytes to the contract schema

limemloh:
## Purpose

Support LEB128 signed and unsigned integers and bytes in the contract schema.

## Changes

- Extended the contract schema with ULeb128, ILeb128, ByteList and ByteArray.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## Unreleased changes
 
-- Extend schema type with ULeb128, ILeb128, ByteList and ByteArray.
-- Add new schema module which include the versioning in the serialization.
-- Change SchemaType implementation for byte arrays.
+- Extend schema type with `ULeb128`, `ILeb128`, `ByteList` and `ByteArray`.
+- Add new schema version which include the versioning in the serialization.
+- Use `schema::Type::ByteList` for `[u8]` implementation of `SchemaType`.
 
 ## concordium-contracts-common 3.0.0 (2022-05-17)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased changes
 
+- Extend schema type with ULeb128, ILeb128, ByteList and ByteArray.
+- Add new schema module which include the versioning in the serialization.
+- Change SchemaType implementation for byte arrays.
+
 ## concordium-contracts-common 3.0.0 (2022-05-17)
 
 - Introduce Entrypoint and Parameter types, and their owned versions.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Unreleased changes
 
 - Extend schema type with `ULeb128`, `ILeb128`, `ByteList` and `ByteArray`.
+  - `ULeb128` and `ILeb128` allow for integers of arbitrary size and are represented in JSON as a string containing the integer.
+  - `ByteList` and `ByteArray` are byte specialized versions of `List` and `Array` and are represented in JSON as lowercase hex encoded strings.  
 - Add new schema version which include the versioning in the serialization.
 - Use `schema::Type::ByteList` for `[u8]` implementation of `SchemaType`.
 
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -934,7 +934,8 @@ mod impls {
                 }
                 Type::ByteList(size_len) => {
                     let len = deserial_length(source, *size_len)?;
-                    let mut string = String::with_capacity(len);
+                    let mut string =
+                        String::with_capacity(std::cmp::min(MAX_PREALLOCATED_CAPACITY, 2 * len));
                     for _ in 0..len {
                         let byte = source.read_u8()?;
                         string.push_str(&format!("{:02x?}", byte));
@@ -943,7 +944,8 @@ mod impls {
                 }
                 Type::ByteArray(len) => {
                     let len = usize::try_from(*len)?;
-                    let mut string = String::with_capacity(len);
+                    let mut string =
+                        String::with_capacity(std::cmp::min(MAX_PREALLOCATED_CAPACITY, 2 * len));
                     for _ in 0..len {
                         let byte = source.read_u8()?;
                         string.push_str(&format!("{:02x?}", byte));
--- a/src/schema.rs
+++ b/src/schema.rs
@@ -45,11 +45,10 @@ pub struct ModuleV1 {
     pub contracts: BTreeMap<String, ContractV1>,
 }
 
-/// Contains all the contract schemas for a module newer than V1 module.
+/// Represents every schema module starting from V1.
 #[derive(Debug, Clone)]
-pub enum VersionedModule {
-    /// Versioned schema module version 0
-    V0(ModuleV1),
+pub enum VersionedModuleSchema {
+    V1(ModuleV1),
 }
 
 /// Describes all the schemas of a V0 smart contract.
@@ -164,9 +163,16 @@ pub enum Type {
     String(SizeLength),
     ContractName(SizeLength),
     ReceiveName(SizeLength),
+    /// An unsigned integer encoded using LEB128 with the addition of a
+    /// constraint on the maximum number of bytes to use for an encoding.
     ULeb128(u32),
+    /// A signed integer encoded using LEB128 with the addition of a constraint
+    /// on the maximum number of bytes to use for an encoding.
     ILeb128(u32),
+    /// A list of bytes. It is serialized with the length first followed by the
+    /// bytes.
     ByteList(SizeLength),
+    /// A fixed sized list of bytes.
     ByteArray(u32),
 }
 
@@ -340,11 +346,11 @@ impl Serial for ModuleV1 {
     }
 }
 
-impl Serial for VersionedModule {
+impl Serial for VersionedModuleSchema {
     fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
         match self {
-            VersionedModule::V0(module) => {
-                out.write_u8(0)?;
+            VersionedModuleSchema::V1(module) => {
+                out.write_u8(1)?;
                 module.serial(out)?;
             }
         }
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased changes
 
+- Extend schema type with ULeb128, ILeb128, ByteList and ByteArray.
+- Add new schema module which include the versioning in the serialization.
```
abizjak, line 6:
```suggestion
- Add new schema version which include the versioning in the serialization.
```
?

src/schema.rs
```diff
@@ -45,6 +45,13 @@ pub struct ModuleV1 {
     pub contracts: BTreeMap<String, ContractV1>,
 }
 
+/// Contains all the contract schemas for a module newer than V1 module.
+#[derive(Debug, Clone)]
+pub enum VersionedModule {
+    /// Versioned schema module version 0
```
abizjak, line 51: This is a confusing comment. I guess it has nothing to do with V0 contracts.

It's also very confusing that your constructor is V0, but the type inside it is called ModuleV1. Why not use `V1`?

src/schema.rs
```diff
@@ -45,6 +45,13 @@ pub struct ModuleV1 {
     pub contracts: BTreeMap<String, ContractV1>,
 }
 
+/// Contains all the contract schemas for a module newer than V1 module.
+#[derive(Debug, Clone)]
+pub enum VersionedModule {
```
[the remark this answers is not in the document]
abizjak: We already have a versioned module elsewhere.

src/schema.rs
```diff
@@ -157,6 +164,10 @@ pub enum Type {
     String(SizeLength),
     ContractName(SizeLength),
     ReceiveName(SizeLength),
+    ULeb128(u32),
+    ILeb128(u32),
+    ByteList(SizeLength),
+    ByteArray(u32),
```
abizjak, line 170: It would be useful to start adding some documentation here. The names are getting very cryptic.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+- Change SchemaType implementation for byte arrays.
```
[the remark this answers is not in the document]
abizjak: You should probably be more precise in saying that there is a new schema type for representing `[u8]` and we change the schema type for `[u8]` to use it.

src/schema.rs
```diff
@@ -845,8 +918,67 @@ mod impls {
...
+                    let int = deserial_biguint(source, *constraint)?;
+                    Ok(Value::String(int.to_string()))
+                }
+                Type::ILeb128(constraint) => {
+                    let int = deserial_bigint(source, *constraint)?;
+                    Ok(Value::String(int.to_string()))
+                }
+                Type::ByteList(size_len) => {
+                    let len = deserial_length(source, *size_len)?;
+                    let mut string = String::new();
```
abizjak, line 931: Use `with_capacity`. I assume the input here is trusted? Otherwise use something like `safe_with_capacity`.

src/schema.rs
```diff
@@ -845,8 +918,67 @@ mod impls {
...
+                    for _ in 0..len {
+                        let byte = source.read_u8()?;
+                        string.push_str(&format!("{:02X?}", byte));
+                    }
+                    Ok(Value::String(string))
+                }
+                Type::ByteArray(len) => {
+                    let len = (*len).try_into()?;
```
abizjak, line 939: What is this for? What is the type of `len` you are assigning to?

src/schema.rs
```diff
@@ -845,8 +918,67 @@ mod impls {
...
+            let value_byte = BigInt::from(byte & 0b0111_1111);
+            let shift = i * 7;
+            result += value_byte << shift;
+
+            if byte & 0b1000_0000 == 0 {
+                if byte & 0b0100_0000 != 0 {
+                    result |= !BigInt::zero() << shift;
+                }
+                return Ok(result);
             }
```
abizjak, line 979: I think these two are sufficiently complex that they should have unit tests.

src/schema.rs
limemloh: I cannot find any documentation for a `safe_with_capacity` function or something like it.
Whould you mind sharing a link?

src/schema.rs
limemloh: Well, my thought was that this is the first version of the _versioned_ schema module, but I see how that is needlessly confusing.
I'll change it

src/schema.rs
abizjak: We have one defined in either wasm-chain-integration, concordium-contracts-common, or crypto-common.

It is just with_capacity, it just has a maximum capacity it preallocates to avoid denial of service with malicious input.

src/schema.rs
abizjak: Great success :)

src/schema.rs
limemloh: Not sure what you are asking.
`len` is used by the for loop, but now also to preallocate the string

abizjak: Yes, but I did not undersand why you did the type conversion 
```
let len = (*len).try_into()?;
```
instead of just using `*len` in the loop bound.

It was unclear to me what the type of the new len was either.

I see now you changed it to `usize::try_from` which is clear. It also makes sense now since the new len is used in the with_capacity function which needs a usize.

src/schema.rs
abizjak: Yeah, too many versions of different things at this point :)

src/schema.rs
limemloh: I have added comments

src/schema.rs
limemloh: I use a safe with capacity now

src/schema.rs
```diff
@@ -157,6 +163,17 @@ pub enum Type {
     String(SizeLength),
     ContractName(SizeLength),
     ReceiveName(SizeLength),
+    /// An unsigned integer encoded using LEB128 with the addition of a
```
Bargsteen, line 166: Consider adding doc comments to all the variants of `Type` now that you've started ;)

src/schema.rs
```diff
@@ -328,6 +346,18 @@ impl Serial for ModuleV1 {
     }
 }
 
+impl Serial for VersionedModuleSchema {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
+        match self {
+            VersionedModuleSchema::V1(module) => {
```
Bargsteen, line 352: So we have three schemas, named and ordered as:
```
schema v1
schema v2
versioned schema v1
```
Is that correct? 
We should add a comment about that, as the reuse of v1 is confusing without knowing the history.

src/schema.rs
```diff
@@ -845,8 +924,209 @@ mod impls {
...
+                    result -= BigInt::from(2).pow(shift)
+                }
+                return Ok(result);
             }
         }
+        Err(ParseError {})
+    }
+
+    #[cfg(test)]
+    mod tests {
```
Bargsteen, line 995: Nice, that you added the tests 👏 🎉

Bargsteen (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased changes
 
+- Extend schema type with `ULeb128`, `ILeb128`, `ByteList` and `ByteArray`.
+- Add new schema version which include the versioning in the serialization.
+- Use `schema::Type::ByteList` for `[u8]` implementation of `SchemaType`.
```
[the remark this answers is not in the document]
limemloh: I have added more about the JSON representation of the new schema types

src/schema.rs
limemloh: well, we have a schema for version 0 smart contracts, which can be found in custom section `concordium-schema-v1`.
We have two schemas for version 1 smart contracts, one in the section `concordium-schema-v2` without version information and one in `concordium-schema` with version information.

Technically schema type also have two versions for v1 smart contracts, since we have extended the type, but this is not tracked in the code

limemloh: I have added a bit more information on the types, but I am not sure how to convey all of this

src/schema.rs
```diff
@@ -33,18 +33,35 @@ pub trait SchemaType {
...
 
+/// Represents every schema module starting from smart contract module V1.
+///
+/// The serialization of this type includes the versioning information.
+///
+/// When embedded into a smart contract module, name the custom section
+/// `concordium-schema`.
+#[derive(Debug, Clone)]
+pub enum VersionedModuleSchema {
+    V1(ModuleV1),
```
abizjak, line 62: So in principle we could also support this for V0 modules. It would have pleasing symmetry.

abizjak: Then the situation isn't that confusing. V0 modules can have custom sections concordium-schema-v1 and concordium-schema, V1 modules can have concordium-schema-v2 and concordium-schema.

The latter one contains the versioning inside, and there are some restrictions on which schema can apply to which module version.

abizjak (approved):

src/schema.rs
limemloh: This is now implemented

src/schema.rs
```diff
@@ -33,18 +33,40 @@ pub trait SchemaType {
...
+/// section labelled `concordium-schema-v2`.
 #[derive(Debug, Clone)]
 pub struct ModuleV1 {
     pub contracts: BTreeMap<String, ContractV1>,
 }
 
+/// Represents the different schema versions.
+///
+/// The serialization of this type includes the versioning information. The
+/// serializtion of this is always prefixed with two 255u8 in order to
```
abizjak, line 57:
```suggestion
/// serialization of this is always prefixed with two 255u8 in order to
```

src/schema.rs
```diff
@@ -328,6 +397,24 @@ impl Serial for ModuleV1 {
     }
 }
 
+impl Serial for VersionedModuleSchema {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
+        // Prefix for versioned module schema, used to distinquish from the unversioned.
+        out.write_u16(u16::MAX)?;
```
abizjak, line 403: This is a nitpick, but I would probably prefer if you were more explicit and wrote
```
out.write_all(&[255, 255])?
```

abizjak: But it is also fine the way it is now.

abizjak (approved):

Bargsteen (approved):
Looks good!

src/schema.rs
limemloh: I started out doing what you suggested, but later when deserializing a u16 seems a bit more convenient.

abizjak (approved):

Issue #43, opened 2022-06-08
Implement the Error for errors


limemloh:
Currently the errors in this crate does not implement the Error trait, which is expected when using the types outside of smart contracts (like the Rust SDK).

We have to ensure that implementing Error does not impact smart contract code, when building using no_std and if so, it should be moved behind a feature flag.

abizjak:
Related to, and should be done together with https://github.com/Concordium/concordium-rust-smart-contracts/issues/76

PR #46, opened 2022-07-26
Add derive crate with serialization

limemloh:
## Purpose

Expose smart contract serialization macros outside of smart contracts https://github.com/Concordium/concordium-rust-smart-contracts/issues/146.
Should be review together with https://github.com/Concordium/concordium-rust-smart-contracts/pull/147.

## Changes

- Moved `concordium-contracts-common` into a subfolder.
- Introduced a crate for macros `concordium-contracts-common-derive`.
- Copied serialization macros from `concordium-std-derive` with adjustments for support use in Rust SDK.

## Checklist

- [x] My code follows the style of this project.
- [x] The code compiles without warnings.
- [x] I have performed a self-review of the changes.
- [ ] I have documented my code, in particular the intent of the
      hard-to-understand areas.
- [x] (If necessary) I have updated the CHANGELOG.

PR #47, opened 2022-07-29
Implement error for error types

limemloh:
## Purpose

Builds on top of https://github.com/Concordium/concordium-contracts-common/pull/46.

Closes https://github.com/Concordium/concordium-contracts-common/issues/43.

## Changes

- Implement `std::error::Error` when the `std` feature is enabled

When compared to https://github.com/Concordium/concordium-contracts-common/pull/46: none of the examples change in size.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-contracts-common/src/types.rs
+++ b/concordium-contracts-common/src/types.rs
@@ -78,6 +78,9 @@ impl fmt::Display for AmountParseError {
     }
 }
 
+#[cfg(feature = "std")]
+impl std::error::Error for AmountParseError {}
+
 /// Parse from string in CCD units. The input string must be of the form
 /// `n[.m]` where `n` and `m` are both digits. The notation `[.m]` indicates
 /// that that part is optional.
@@ -394,6 +397,9 @@ impl fmt::Display for ParseTimestampError {
 }
 
 #[cfg(feature = "derive-serde")]
+impl std::error::Error for ParseTimestampError {}
+
+#[cfg(feature = "derive-serde")]
 /// The FromStr parses the time according to RFC3339.
 impl str::FromStr for Timestamp {
     type Err = ParseTimestampError;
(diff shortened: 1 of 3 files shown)


concordium-contracts-common/src/types.rs
```diff
@@ -78,6 +78,9 @@ impl fmt::Display for AmountParseError {
     }
 }
 
+#[cfg(feature = "std")]
+impl std::error::Error for AmountParseError {}
```
abizjak, line 82: So this is a bit wonky. It uses the default implementation which will have `.description` return "Deprecated, use display".

Is this intended?

concordium-contracts-common/src/types.rs
```diff
@@ -393,6 +396,9 @@ impl fmt::Display for ParseTimestampError {
     }
 }
 
+#[cfg(feature = "derive-serde")]
+impl std::error::Error for ParseTimestampError {}
```
abizjak, line 400: Likewise here.

concordium-contracts-common/src/types.rs
```diff
@@ -1310,6 +1325,13 @@ pub struct ParseError {}
 /// of parsing from binary via the `Serial` instance.
 pub type ParseResult<A> = Result<A, ParseError>;
 
+impl fmt::Display for ParseError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result { write!(f, "Parsing failed") }
```
abizjak, line 1329: Maybe better to use the more precise `f.write_str` in this case.

abizjak (commented):
Could we maybe use `thiserror` in this crate to derive Display + Error types?

concordium-contracts-common/src/types.rs
limemloh: That is the current intended way of implementing this, and seems to also be how the `thiserror::Error` derive macro behaves.

abizjak (approved):

PR #49, opened 2022-08-10
More ergonomic policies

Bargsteen:
## Purpose

Part of https://github.com/Concordium/concordium-rust-smart-contracts/pull/144

## Changes

- Change `AttributeValue` from slice to struct.
   - Add init methods, serialization, and `From` for 0..31 sized bytearrays via macro.
- Remove `OwnedAttributeValue`.

## Checklist

      hard-to-understand areas.


concordium-contracts-common/src/impls.rs
```diff
@@ -980,6 +980,26 @@ impl Deserial for AttributeTag {
...
+
+impl Deserial for AttributeValue {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        let mut buf = [0u8; 32];
+        let len: u8 = source.get()?;
+        buf[0] = len;
+        if len > 31 {
+            return Err(ParseError::default());
+        }
+        source.read_exact(&mut buf[1..len as usize + 1])?;
```
abizjak, line 998:
```suggestion
        source.read_exact(&mut buf[1..=len as usize])?;
```

concordium-contracts-common/src/types.rs
```diff
@@ -1081,25 +1081,142 @@ pub struct Cursor<T> {
     pub data:   T,
 }
 
+/// Errors that can occur when constructing a new [`AttributeValue`].
+#[derive(Debug, PartialEq, Eq)]
+pub enum NewAttributeValueError {
```
abizjak, line 1086: Shouldn't we derive/implement `Error` for this as well?

abizjak: Assuming `std` feature.

concordium-contracts-common/src/types.rs
```diff
@@ -1081,25 +1081,142 @@ pub struct Cursor<T> {
...
+macro_rules! from_bytearray_to_attribute_value {
+    ($n:expr) => {
+        impl From<[u8; $n]> for AttributeValue {
+            fn from(data: [u8; $n]) -> Self { AttributeValue::new(&data[..]).unwrap() }
+        }
+
+        impl From<&[u8; $n]> for AttributeValue {
+            fn from(data: &[u8; $n]) -> Self { AttributeValue::new(&data[..]).unwrap() }
+        }
+    };
```
abizjak, line 1176: I'm not sure this is useful. `[u8; N]` is Copy.

concordium-contracts-common/src/types.rs
```diff
@@ -1081,25 +1081,149 @@ pub struct Cursor<T> {
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[repr(transparent)]
+pub struct AttributeValue {
+    pub(crate) inner: [u8; 32],
+}
+
+impl AttributeValue {
+    /// Create a new [`Self`] from a slice of bytes. The slice must have a
+    /// length of *at most 31 bytes*.
+    pub fn new(data: &[u8]) -> Result<Self, NewAttributeValueError> {
```
limemloh, line 1123: Minor thing, I find it a bit inconsistent to use a slice here, when `new_unchecked` does not.
I would prefer if we had a `new(data, [u8;32])` and a `try_from<&[u8]>`

limemloh (dismissed):

concordium-contracts-common/src/types.rs
Bargsteen: I decided to keep the `new` method because:

1. It is the easiest one to find for users.
2. The slice allows you to provide just the bytes you need instead of an array with padded zeroes (and a length).
3. We can't check that the length encoded in the first byte is correct (the sequence might end with one or more zeroes).

Instead, I added a `TryFrom` instance, which also makes sense to have :blush:.

abizjak (approved):

abizjak:
There are some conflicts in the changelog, so some care is needed there.

abizjak (approved):

PR #52, opened 2022-08-19
Add v2 contract schemas

Bargsteen:
## Purpose

Adds a v2 schema, which supports schemas for errors.
Part of the solution for https://github.com/Concordium/concordium-rust-smart-contracts/issues/155

## Changes

- Rename some types by adding a `V1` suffix.
- Adding some schema `V2` types and implementing serialization and helpers.

## Checklist

- [x ] I have documented my code, in particular the intent of the
      hard-to-understand areas.

### DIFF
--- a/concordium-contracts-common/CHANGELOG.md
+++ b/concordium-contracts-common/CHANGELOG.md
@@ -16,6 +16,7 @@
 - Rework attribute values types (breaking change)
   - Change `AttributeValue` from a slice to a struct.
   - Remove `OwnedAttributeValue` type.
+- Add support for smart contract v2 schemas.
 
 ## concordium-contracts-common 3.1.0 (2022-08-04)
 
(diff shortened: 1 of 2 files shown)


abizjak (approved):
In hindsight it would probably be better to just use a struct with `Option<...>` fields.

We would only have the one wart with all `None` fields, but less annoying code.

But that ship has sailed. This looks good to me. Some serialization tests would be nice though.

Bargsteen:
> In hindsight it would probably be better to just use a struct with `Option<...>` fields.
> 
> We would only have the one wart with all `None` fields, but less annoying code.
> 
> But that ship has sailed. This looks good to me. Some serialization tests would be nice though.

I did consider that option, but I decided against it due to the three `None`s and the two extra bytes needed for encoding it (unless we used some serialization trick).

PR #54, opened 2022-10-10
Add schema version3

### LINKED ISSUE #53: Event schema
Task description

Update concordium-contract-common to support event schema creation/serialization/deserialization.

DOBEN:
## Purpose

closes #53 

Add schemas for events

## Changes

Add schema version3

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-contracts-common/CHANGELOG.md
+++ b/concordium-contracts-common/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 ## Unreleased changes
 
-## concordium-contracts-common 4.1.0 (Unreleased)
-
 - Add support for smart contract v3 schemas.
 - Add type `ModuleReference` representing a module reference.
 - Implement `SchemaType` for `OwnedEntrypointName` and `OwnedParameter`.
--- a/concordium-contracts-common/CHANGELOG.md
+++ b/concordium-contracts-common/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased changes
 
+## concordium-contracts-common 4.1.0 (Unreleased)
+
+- Add support for smart contract v3 schemas.
 - Add type `ModuleReference` representing a module reference.
 - Implement `SchemaType` for `OwnedEntrypointName` and `OwnedParameter`.
 
@@ -27,7 +30,7 @@
 
 - Extend schema type with `ULeb128`, `ILeb128`, `ByteList` and `ByteArray`.
   - `ULeb128` and `ILeb128` allow for integers of arbitrary size and are represented in JSON as a string containing the integer.
-  - `ByteList` and `ByteArray` are byte specialized versions of `List` and `Array` and are represented in JSON as lowercase hex encoded strings.  
+  - `ByteList` and `ByteArray` are byte specialized versions of `List` and `Array` and are represented in JSON as lowercase hex encoded strings.
 - Add new schema version which include the versioning in the serialization.
 - Use `schema::Type::ByteList` for `[u8]` implementation of `SchemaType`.
 - Introduce `HasSize` trait.
@@ -39,7 +42,7 @@
 ## concordium-contracts-common 3.0.0 (2022-05-17)
 
 - Introduce Entrypoint and Parameter types, and their owned versions.
-- Add a new schema version for V1 smart contracts. 
+- Add a new schema version for V1 smart contracts.
   This adds schema for return values of init and receive functions, and removes the state schema.
 - `get_chain_name`, `get_name_parts` have been moved from `OwnedReceiveName` to
   `ReceiveName`. The method `get_func_name` of `OwnedReceiveName` became
--- a/concordium-contracts-common/src/schema.rs
+++ b/concordium-contracts-common/src/schema.rs
@@ -60,6 +60,7 @@ pub struct ModuleV2 {
 
 /// Contains all the contract schemas for a smart contract module V1 with a V3
 /// schema.
+#[cfg_attr(test, derive(arbitrary::Arbitrary))]
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ModuleV3 {
     pub contracts: BTreeMap<String, ContractV3>,
@@ -111,6 +112,7 @@ pub struct ContractV2 {
 }
 
 /// Describes all the schemas of a V1 smart contract with a V3 schema.
+#[cfg_attr(test, derive(arbitrary::Arbitrary))]
 #[derive(Debug, Default, Clone, PartialEq, Eq)]
 /// The [`Default`] instance produces an empty schema.
 pub struct ContractV3 {
@@ -165,6 +167,7 @@ impl FunctionV1 {
 /// Describes the schema of an init or a receive function for V1 contracts with
 /// V3 schemas. Differs from [`FunctionV1`] in that a schema for the error can
 /// be included.
+#[cfg_attr(test, derive(arbitrary::Arbitrary))]
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct FunctionV2 {
     pub parameter:    Option<Type>,
@@ -711,107 +714,59 @@ impl Serial for FunctionV2 {
         let parameter = self.parameter.is_some();
         let return_value = self.return_value.is_some();
         let error = self.error.is_some();
-        match (parameter, return_value, error) {
+        let tag: u8 = match (parameter, return_value, error) {
             // parameter
-            (true, false, false) => {
-                out.write_u8(0)?;
-                self.parameter.serial(out)
-            }
+            (true, false, false) => 0,
             // return_value
-            (false, true, false) => {
-                out.write_u8(1)?;
-                self.return_value.serial(out)
-            }
+            (false, true, false) => 1,
             // parameter + return_value
-            (true, true, false) => {
-                out.write_u8(2)?;
-                self.parameter.serial(out)?;
-                self.return_value.serial(out)
-            }
+            (true, true, false) => 2,
             // error
-            (false, false, true) => {
-                out.write_u8(3)?;
-                self.error.serial(out)
-            }
+            (false, false, true) => 3,
             // parameter + error
-            (true, false, true) => {
-                out.write_u8(4)?;
-                self.parameter.serial(out)?;
-                self.error.serial(out)
-            }
+            (true, false, true) => 4,
             // return_value + error
-            (false, true, true) => {
-                out.write_u8(5)?;
-                self.return_value.serial(out)?;
-                self.error.serial(out)
-            }
+            (false, true, true) => 5,
             // parameter + return_value + error
-            (true, true, true) => {
-                out.write_u8(6)?;
-                self.parameter.serial(out)?;
-                self.return_value.serial(out)?;
-                self.error.serial(out)
-            }
+            (true, true, true) => 6,
             // no schema: cannot happen
-            (false, false, false) => out.write_u8(7),
+            (false, false, false) => 7,
+        };
+        out.write_u8(tag)?;
+        if let Some(p) = self.parameter.as_ref() {
+            p.serial(out)?;
         }
+        if let Some(rv) = self.return_value.as_ref() {
+            rv.serial(out)?;
+        }
+        if let Some(err) = self.error.as_ref() {
+            err.serial(out)?;
+        }
+        Ok(())
     }
 }
 
 impl Deserial for FunctionV2 {
     fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
         let idx = source.read_u8()?;
-        match idx {
-            // parameter
-            0 => Ok(FunctionV2 {
-                parameter:    source.get()?,
-                return_value: None,
-                error:        None,
-            }),
-            // return_value
-            1 => Ok(FunctionV2 {
-                parameter:    None,
-                return_value: source.get()?,
-                error:        None,
-            }),
-            // parameter + return_value
-            2 => Ok(FunctionV2 {
-                parameter:    source.get()?,
-                return_value: source.get()?,
-                error:        None,
-            }),
-            // error
-            3 => Ok(FunctionV2 {
-                parameter:    None,
-                return_value: None,
-                error:        source.get()?,
-            }),
-            // parameter + error
-            4 => Ok(FunctionV2 {
-                parameter:    source.get()?,
-                return_value: None,
-                error:        source.get()?,
-            }),
-            // return_value + error
-            5 => Ok(FunctionV2 {
-                parameter:    None,
-                return_value: source.get()?,
-                error:        source.get()?,
-            }),
-            // parameter + return_value + error
-            6 => Ok(FunctionV2 {
-                parameter:    source.get()?,
-                return_value: source.get()?,
-                error:        source.get()?,
-            }),
-            // no schema
-            7 => Ok(FunctionV2 {
-                parameter:    None,
-                return_value: None,
-                error:        None,
-            }),
-            _ => Err(ParseError::default()),
+        let mut r = FunctionV2 {
+            parameter:    None,
+            return_value: None,
+            error:        None,
+        };
+        if idx > 7 {
+            return Err(ParseError::default());
         }
+        if matches!(idx, 0 | 2 | 4 | 6) {
+            let _ = r.parameter.insert(source.get()?);
+        }
+        if matches!(idx, 1 | 2 | 5 | 6) {
+            let _ = r.return_value.insert(source.get()?);
+        }
+        if matches!(idx, 3 | 4 | 5 | 6) {
+            let _ = r.error.insert(source.get()?);
+        }
+        Ok(r)
     }
 }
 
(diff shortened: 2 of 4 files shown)


concordium-contracts-common/src/schema.rs
```diff
@@ -101,6 +110,45 @@ pub struct ContractV2 {
...
+/// Contains event schema and associated event tag vector.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct EventSchema {
+    /// Event schema.
+    pub event_schema: Type,
+    /// Event tags are mapped to enum indexes:
+    /// The event at index 0 in the event schema enum is mapped to the event tag
+    /// value `event_tags[0]`. The event at index 1 in the event schema enum
+    /// is mapped to the event tag value `event_tags[1]`. ...
+    pub event_tags:   Option<Vec<u8>>,
```
abizjak, line 130: I am not convinced it is a good idea to add these tags to the EventSchema.

In my view we should add a new constructor/variant to the `Schema::Type` enum that allows you to supply custom tags for your enum.

these tags can be either supplied via the derive annotation using an extra attribute, e.g., `tag =`, or via a manual implementation of `SchemaType` for the given type.

Would there be a problem with this?

We are in any case adding a new schema version.

concordium-contracts-common/src/schema.rs
```diff
@@ -229,6 +277,214 @@ impl FunctionV2 {
...
+    },
+    ErrorEvent {
+        error: Type,
+        event: EventSchema,
+    },
+    ParamErrorEvent {
+        parameter: Type,
+        error:     Type,
+        event:     EventSchema,
+    },
```
abizjak, line 327: I think this is getting out of hand. I strongly recommend using instead

```
pub struct FunctionV3 {
   parameter: Option<Type>,
   return_value: Option<Type>,
   error: OPtion<Type>,
   event: Option<Type>
}
```
instead. And then have the property that at least one of them is not None (which we can check during deserialization).

concordium-contracts-common/src/schema.rs
```diff
@@ -780,6 +1081,192 @@ impl Deserial for FunctionV2 {
...
+                parameter.serial(out)
+            }
+            FunctionV3::Rv(return_value) => {
+                out.write_u8(1)?;
+                return_value.serial(out)
+            }
+            FunctionV3::ParamRv {
+                parameter,
+                return_value,
+            } => {
```
abizjak, line 1098: If you use the struct, as I suggest you do, then you can use a single `u8` tag for encoding which parameters are `Some` (e.g., bit 0 means parameter is Some, bit 1 set means return value is Some, ...)

concordium-contracts-common/src/schema.rs
```diff
@@ -1213,6 +1411,24 @@ mod impls {
...
+                            .ok_or_else(ParseError::default)?
+                    } else {
+                        // Event schemas are not supported for more than 256 event variants.
+                        return Err(ParseError::default());
+                    };
+
+                    let (name, fields_ty, tag) =
+                        variants.get(enum_index).ok_or_else(ParseError::default)?;
+                    let fields = fields_ty.to_json(source)?;
+                    Ok(json!({"EventLogIndex": tag, name: fields}))
```
abizjak, line 1430: Is it necessary to have this `EventLogIndex` in the JSON representation? I would have thought that the name would give the index (assuming we have the schema).

abizjak (commented):
Looks great in general.

I have a few specific comments where we might be able to simplify things a little bit more.

concordium-contracts-common/src/schema.rs
```diff
@@ -101,6 +110,32 @@ pub struct ContractV2 {
     pub receive: BTreeMap<String, FunctionV2>,
 }
 
+/// Describes all the schemas of a V1 smart contract with a V3 schema.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+/// The [Default] instance produces an empty schema.
+pub struct ContractV3 {
+    pub init:    Option<FunctionV3>,
+    pub receive: BTreeMap<String, FunctionV3>,
```
abizjak, line 118: This means that each function can also have an event schema. What is meant to be the meaning of this if both the init function has the event schema, and the receive function has it?

Was not the idea that the contract itself has an event schema, not the individual entrypoints. In that case, would not 
```
pub struct ContractV3 {
   pub init: ...
   pub receive: ...
   pub event: Option<Type>
}
```
be the more appropriate modelling of that?

concordium-contracts-common/src/schema.rs
```diff
@@ -1213,6 +1411,24 @@ mod impls {
...
+                            .iter()
+                            .position(|item| item.2 == idx)
+                            .ok_or_else(ParseError::default)?
+                    } else {
+                        // Event schemas are not supported for more than 256 event variants.
+                        return Err(ParseError::default());
+                    };
+
+                    let (name, fields_ty, tag) =
+                        variants.get(enum_index).ok_or_else(ParseError::default)?;
```
limemloh, line 1428: You could use [`find`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find) here, instead of first finding the index to lookup with.
Something like this:
```suggestion
                        variants.find(|item| item.2 == idx).ok_or_else(ParseError::default)?;
```

concordium-contracts-common/src/schema.rs
limemloh: I agree, the tag itself is a detail which we can hide from the JSON representation.
The name of the variant is more useful for the user.

concordium-contracts-common/src/schema.rs
```diff
@@ -1507,6 +1723,23 @@ mod tests {
         assert_eq!(serial_deserial(&f6), Ok(f6));
     }
 
+    #[test]
+    fn test_function_v3_serial_deserial_is_id() {
+        let f1 = FunctionV3 {
+            index:        15,
```
limemloh, line 1729: I think you could remove `index` from the struct, but if not it would be easier to read:
```suggestion
            index:        0b1111,
```

concordium-contracts-common/src/schema.rs
```diff
@@ -229,6 +264,32 @@ impl FunctionV2 {
     }
 }
 
+/// Describes the schema of an init or a receive function for V1 contracts with
+/// V3 schemas. Differs from [`FunctionV2`] in that a schema for the events can
+/// be included.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FunctionV3 {
+    pub index:        u8,
```
limemloh, line 272: Is this field ever needed?
It seems redundant, since the other fields defines its value

limemloh (changes requested):

concordium-contracts-common/src/schema.rs
```diff
@@ -1213,6 +1411,24 @@ mod impls {
...
                     Ok(json!({ name: fields }))
                 }
+                Type::EnumEvent(variants) => {
+                    let enum_index = if variants.len() <= 256 {
+                        let idx = u8::deserial(source)?;
+                        // Find the enum index which is associated to the event tag.
+                        variants
+                            .iter()
+                            .position(|item| item.2 == idx)
+                            .ok_or_else(ParseError::default)?
```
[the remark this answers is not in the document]
abizjak: But that is relevant because there are not specific tags. Here the tags determine the order, so a map would make sense for this.

concordium-contracts-common/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased changes
 
+## concordium-contracts-common 4.1.0 (Unreleased)
```
abizjak, line 5: This should still be under "Unreleased changes". The process is that I will tag this with the version when I do a release on crates.io.

concordium-contracts-common/src/schema.rs
```diff
@@ -229,6 +252,27 @@ impl FunctionV2 {
     }
 }
 
+/// Describes the schema of an init or a receive function for V1 contracts with
+/// V3 schemas. Differs from [`FunctionV2`] in that a schema for the events can
+/// be included.
```
abizjak, line 257: I think the second sentence should be removed now. It is no longer true. We could also just have FunctionV2 instead of this type seeing that they are the same. And then use FunctionV2 in the `init` and `receive` parts of `ContractV3`.

abizjak (commented):
Generally good, however I would recommend removing FunctionV3 since it is now equal to FunctionV2.

That will remove a bunch more code :)

concordium-contracts-common/src/schema.rs
```diff
@@ -780,6 +874,76 @@ impl Deserial for FunctionV2 {
     }
 }
 
+impl Serial for FunctionV3 {
```
Bargsteen: If we want to just use FunctionV2 for both V2 and V3 contract schemas, and we want the new and improved 3 `Option`s version, then we need to make it compatible with the existing v2 serialization. I.e.:

0 => param
1 => rv
 2 => param + rv
3 => error
4 => param + error
5 => rv + error
6 => param + rv + error

vikt0r0:
Not sure if it's a coincidence, but I get "Failed reading: Invalid FunctionSchemaV2 tag: 7", perhaps corresponding to [line 752 in schema.rs](https://github.com/Concordium/concordium-contracts-common/pull/54/commits/0fb6ee1ea4b842c9d18a95eb188132fde4360357#diff-b621ae2a4dcb0ed2f581f90102e8dc2e7714eef78b25e978c0e5d85b2c2b9b9a) when inspecting this SC in concordium-client. The contract has index 1013 on stagenet.

concordium-contracts-common/src/schema.rs
```diff
@@ -700,53 +704,52 @@ impl Deserial for FunctionV1 {
...
-                error.serial(out)
+                self.parameter.serial(out)?;
+                self.return_value.serial(out)?;
+                self.error.serial(out)
             }
+            // no schema: cannot happen
+            (false, false, false) => out.write_u8(7),
         }
     }
 }
```
abizjak, line 755: Are you sure this is correct. You seem to be serializing `Option<Type>` all the time, which as far as I understand is not the same as it used to be.

I think what you want is
```rust

impl Serial for FunctionV2 {
    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
        let parameter = self.parameter.is_some();
        let return_value = self.return_value.is_some();
        let error = self.error.is_some();
        let tag: u8 = match (parameter, return_value, error) {
            // parameter
            (true, false, false) => 0,
            // return_value
            (false, true, false) => 1,
            // parameter + return_value
            (true, true, false) => 2,
            // error
            (false, false, true) => 3,
            // parameter + error
            (true, false, true) => 4,
            // return_value + error
            (false, true, true) => 5,
            // parameter + return_value + error
            (true, true, true) => 6,
            // no schema: cannot happen
            (false, false, false) => 7,
        };
        out.write_u8(tag)?;
        if let Some(p) = self.parameter.as_ref() {
            p.serial(out)?;
        }
        if let Some(rv) = self.return_value.as_ref() {
            rv.serial(out)?;
        }
        if let Some(err) = self.error.as_ref() {
            err.serial(out)?;
        }
        Ok(())
    }
}
```

Or am I misunderstanding something?

concordium-contracts-common/src/schema.rs
```diff
@@ -755,26 +758,54 @@ impl Deserial for FunctionV2 {
     fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
         let idx = source.read_u8()?;
         match idx {
-            0 => Ok(FunctionV2::Param(source.get()?)),
-            1 => Ok(FunctionV2::Rv(source.get()?)),
-            2 => Ok(FunctionV2::ParamRv {
+            // parameter
+            0 => Ok(FunctionV2 {
                 parameter:    source.get()?,
```
abizjak, line 763:
```suggestion
                parameter:    Some(source.get()?),
```

concordium-contracts-common/src/schema.rs
```diff
@@ -755,26 +758,54 @@ impl Deserial for FunctionV2 {
...
+                return_value: None,
+                error:        None,
+            }),
+            // return_value
+            1 => Ok(FunctionV2 {
+                parameter:    None,
                 return_value: source.get()?,
```
abizjak, line 770:
```suggestion
                return_value: Some(source.get()?),
```

concordium-contracts-common/src/schema.rs
```diff
@@ -755,26 +758,54 @@ impl Deserial for FunctionV2 {
...
+                error:        None,
             }),
-            3 => Ok(FunctionV2::Error(source.get()?)),
-            4 => Ok(FunctionV2::ParamError {
-                parameter: source.get()?,
-                error:     source.get()?,
+            // parameter + return_value
+            2 => Ok(FunctionV2 {
+                parameter:    source.get()?,
+                return_value: source.get()?,
```
abizjak, line 776:
```suggestion
                parameter:    Some(source.get()?),
                return_value: Some(source.get()?),
```

concordium-contracts-common/src/schema.rs
```diff
@@ -755,26 +758,54 @@ impl Deserial for FunctionV2 {
...
             }),
+            // no schema
+            7 => Ok(FunctionV2 {
+                parameter:    None,
+                return_value: None,
+                error:        None,
+            }),
             _ => Err(ParseError::default()),
         }
     }
```
abizjak, line 811: I think this has the same issue as Serial, in that `source.get()` you are actually deserializing `Option<Type>`. But you want `Type` since the tag already tells you the option. I think you want this

```rust

impl Deserial for FunctionV2 {
    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
        let idx = source.read_u8()?;
        let mut r = FunctionV2 {
            parameter:    None,
            return_value: None,
            error:        None,
        };
        if idx > 7 {
            return Err(ParseError::default());
        }
        if matches!(idx, 0 | 2 | 4 | 6) {
            let _ = r.parameter.insert(source.get()?);
        }
        if matches!(idx, 1 | 2 | 5 | 6) {
            let _ = r.return_value.insert(source.get()?);
        }
        if matches!(idx, 3 | 4 | 5 | 6) {
            let _ = r.error.insert(source.get()?);
        }
        Ok(r)
    }
}
```

concordium-contracts-common/src/schema.rs
```diff
@@ -1545,14 +1575,58 @@ mod tests {
...
+                    }),
+                    ("c".into(), FunctionV2 {
+                        parameter:    Some(Type::String(SizeLength::U32)),
+                        return_value: Some(Type::String(SizeLength::U32)),
+                        error:        Some(Type::String(SizeLength::U32)),
                     }),
                 ]),
+                event:   Some(Type::TaggedEnum(event_map)),
             })]),
         };
```
abizjak, line 1631: There does not appear to be any test here. I suggest adding a general randomized test

```rust
    #[test]
    fn test_module_v3_schema_serial_deserial_is_id() {
        use rand::prelude::*;
        use rand_pcg::Pcg64;

        let seed: u64 = random();
        let mut rng = Pcg64::seed_from_u64(seed);
        let mut data = [0u8; 100000];
        rng.fill_bytes(&mut data);

        let mut unstructured = Unstructured::new(&data);

        for _ in 0..10000 {
            let schema = ModuleV3::arbitrary(&mut unstructured).unwrap();

            let res = from_bytes::<ModuleV3>(&to_bytes(&schema)).unwrap();
            assert_eq!(schema, res);
        }
    }
```

You will need to add 
```
#[cfg_attr(test, derive(arbitrary::Arbitrary))]
```
to ModuleV3, ContractV3

abizjak (commented):
I think this looks good overall. I believe there are bugs in serialization with respect to compatibility with the FunctionV2.

I suggested how to fix those. Once that is done I think this is good to go.

Bargsteen (approved):
Nice work ;)

concordium-contracts-common/src/schema.rs
```diff
@@ -333,6 +290,8 @@ pub enum Type {
     ByteList(SizeLength),
     /// A fixed sized list of bytes.
     ByteArray(u32),
+    /// An enum with a tag.
```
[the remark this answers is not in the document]
abizjak: I don't think this type is very odd. In fact the Enum one is more odd, since it hardcodes the tags to be the indices in the list, which is ad-hoc. This is the more "obvious" type.

I would also not say that this is anything to do with events. This can be used as any other type. It makes sense for a parameter as well.

abizjak (approved):

concordium-contracts-common/src/schema.rs
```diff
@@ -704,83 +711,62 @@ impl Deserial for FunctionV1 {
...
+            // error
+            (false, false, true) => 3,
+            // parameter + error
+            (true, false, true) => 4,
+            // return_value + error
+            (false, true, true) => 5,
+            // parameter + return_value + error
+            (true, true, true) => 6,
+            // no schema: cannot happen
+            (false, false, false) => 7,
```
limemloh, line 733: If this case cannot happen, then consider returning an error instead

concordium-contracts-common/src/schema.rs
```diff
@@ -704,83 +711,62 @@ impl Deserial for FunctionV1 {
...
-                return_value: source.get()?,
-                error:        source.get()?,
-            }),
-            _ => Err(ParseError::default()),
+        let mut r = FunctionV2 {
+            parameter:    None,
+            return_value: None,
+            error:        None,
+        };
+        if idx > 7 {
```
limemloh, line 757: I see you still allow for the tag to be 7, so I guess it can happen

concordium-contracts-common/src/schema.rs
```diff
@@ -704,83 +711,62 @@ impl Deserial for FunctionV1 {
...
+            return Err(ParseError::default());
         }
+        if matches!(idx, 0 | 2 | 4 | 6) {
+            let _ = r.parameter.insert(source.get()?);
```
limemloh, line 761: Very minor thing, but I don't think you need the `let _ =` part
```suggestion
            r.parameter.insert(source.get()?);
```

limemloh (approved):
LGTM

concordium-contracts-common/src/schema.rs
DOBEN: I get a warning from `clippy` about an unused return value if I remove it. That is why I kept the initial version now.

concordium-contracts-common/src/schema.rs
DOBEN: I updated the comment:

` // no schema:`

concordium-contracts-common/src/schema.rs
abizjak: Yes, clippy in more recent versions complains about unused option values. So this has to be there :)

concordium-contracts-common/src/schema.rs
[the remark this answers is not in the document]
Bargsteen: I see your point. Okay, it's not that odd then ;)

concordium-contracts-common/src/schema.rs
Bargsteen: The type, with three `Option`s, technically allow this case. But, hopefully, our tools should never construct a `FunctionSchemaV2` with three `None`s. 
We could make the fields private and add a `new` method to check that at least one of the schemas was supplied. But I don't see much benefit in that because it is only used internally and because we probably access the fields directly in several places.
So let's just keep the current version.

PR #57, opened 2022-10-24
Implement SchemaType for OwnedParameter

limemloh:
## Purpose

Implement `SchemaType` for `OwnedParameter`


abizjak (approved):
This does not hurt, so approved, but is also not really necessary for how the schema is being used.

Bargsteen (approved):

PR #58, opened 2022-10-28
Moved a "schema-json" function and added versioned schema helpers

rasmus-kirk:
## Purpose

I found that the `write_bytes_from_json_schema_type()` in [concordium-smart-contract-tools](https://github.com/Concordium/concordium-smart-contract-tools/blob/7a9a571fcc76f084afdd759ef0d4ea99203fa2b6/cargo-concordium/src/schema_json.rs) could be reused in the nodejs sdk, so this is a proposal to move that function to this repo. 

The functions in `versioned_schema_helpers.rs` are based on the code snippets from the [mobile wallet](https://github.com/Concordium/concordium-base/blob/ffae5a6de03aff55d0d5790d94987fb5bb82a717/mobile_wallet/src/lib.rs) lines 197-249 & 269-278. Some more functions with similar use-cases have been added as they will be necessary for an upcoming PR in the ndoejs sdk.

PR's to concordium-base and concordium-smart-contract-tools that remove the code from those repos will follow.

## Changes

The referenced code above have been moved, and the functions `get_return_value_schema()` and `get_init_param_schema()` have been added.

## Checklist

      hard-to-understand areas.


rasmus-kirk:
Made this a PR on the [53-event-schema](https://github.com/Concordium/concordium-contracts-common/tree/53-event-schema) branch because [concordium-smart-contract-tools](https://github.com/Concordium/concordium-smart-contract-tools) needs to be updated to at least this repos main branch. This has been done on the [9-event-schemas](https://github.com/Concordium/concordium-smart-contract-tools/tree/9-event-schemas) branch so I'm thinking we just add on top of that branch when it has been merged.

concordium-contracts-common/Cargo.toml
```diff
@@ -55,11 +55,19 @@ version = "1.0"
 path = "../concordium-contracts-common-derive"
 version = "1"
 
+[dependencies.anyhow]
+optional = true
+version = "1.0"
```
abizjak, line 60: I do not want to bring anyhow to contracts-common. Since the serialization is now part of a library we should define a precise error type for it and use thiserror to derive the relevant things for it.

concordium-contracts-common/src/lib.rs
```diff
@@ -89,7 +89,11 @@ mod traits;
 mod impls;
 pub mod constants;
 pub mod schema;
+#[cfg(feature = "derive-serde")]
+pub mod schema_json;
 mod types;
+#[cfg(feature = "derive-serde")]
+pub mod versioned_schema_helpers;
```
abizjak, line 96: It is hard to use a library that has exports scattered like this.

The `schema_json` should just be part of the `schema` (feature gated of course) module since there is one function there. The `write_bytes_from_json_schema_type` should instead be a method on the `Type`, i.e, 

```
impl Type {
    pub fn write_bytes_from_json_schema(&self, json: &serde_json::Value, out: ...) -> Result<(), JSONConversionError> {
        ....
   }
}

File: concordium-contracts-common/src/schema_json.rs [hunk 3]
```diff
@@ -0,0 +1,636 @@
+use crate::{schema::*, *};
+use anyhow::{anyhow, bail, ensure, Context};
+use num_bigint::{BigInt, BigUint};
+use num_traits::Zero;
+use serde_json::Value;
+use std::convert::TryInto;
+
+macro_rules! serial {
```
abizjak, line 8: DOcument

abizjak (changes requested):

concordium-contracts-common/src/schema_json.rs
```diff
@@ -0,0 +1,728 @@
...
+        $value.serial($out).or(Err(JsonError::FailedWriting))
+    };
+}
+
+// Returns a WrongJsonType error with the provided error message
+macro_rules! wrong_json_type {
+    ($msg:literal) => {
+        Err(JsonError::WrongJsonType($msg.to_string()))
+    };
+}
```
abizjak, line 59: Please remove this macro. THis does not bring value and just obfuscates the code.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -0,0 +1,728 @@
...
+
+// Either the given condition holds otherwise the provided error and
+// accompanying message is returned
+macro_rules! ensure {
+    ($err:ident, $cond:expr, $msg:expr) => {
+        if !$cond {
+            return Err(JsonError::$err($msg.to_string()));
+        }
+    };
+}
```
abizjak, line 69: This macro has odd arguments. The anyhow ensure macro takes the error message as one last argument, which this one should also do. You can bring all JSONError constructors into scope using `use JsonError::*`

concordium-contracts-common/src/schema_json.rs
abizjak: Use the constructor directly. The savings in a few characters are not worth it.

concordium-contracts-common/src/versioned_schema_helpers.rs
```diff
@@ -0,0 +1,316 @@
+use crate::{from_bytes, schema::*, types};
+use thiserror;
+
+enum VersionedContractSchema {
+    V0(ContractV0),
+    V1(ContractV1),
+    V2(ContractV2),
+    V3(ContractV3),
+}
```
abizjak, line 9: Why is this a private type only? Don't we need it elsewhere?

concordium-contracts-common/src/versioned_schema_helpers.rs
```diff
@@ -0,0 +1,316 @@
...
+
+#[derive(Debug, thiserror::Error)]
```
abizjak, line 11: Derive Clone, Copy

concordium-contracts-common/src/lib.rs
abizjak, line 96: This should not be a public module. The only type that is exported from there should be moved to `schema` or `schema_json`.  This is an internal module only and pollutes the public API.

concordium-contracts-common/src/schema_json.rs
rasmus-kirk: Yeah looks better, thanks

abizjak:
Please put this on top of main. There is a conflict now.

Also please make corresponding PRs in 
- concordium-base
- smart-contract-tools

We want to merge them all at the same time to avoid issues. 

Otherwise looks fine.

abizjak:
Also please edit the title of this PR so that it is useful.

PR #59, opened 2022-11-01
Add type for exchange rates

limemloh:
## Purpose

Add `ExchangeRate` type needed for the contract chain queries.

## Changes

- Introduce `ExchangeRate` representing an exchange rate between two quantities.
- Added optional dependency on `num-integer` for the `gcd` function, only when enabling `derive-serde`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-contracts-common/src/impls.rs
+++ b/concordium-contracts-common/src/impls.rs
@@ -163,6 +163,21 @@ impl Deserial for Amount {
     }
 }
 
+impl Serial for ExchangeRate {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
+        out.write_u64(self.numerator())?;
+        out.write_u64(self.denominator())
+    }
+}
+
+impl Deserial for ExchangeRate {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        let numerator = source.read_u64()?;
+        let denominator = source.read_u64()?;
+        Ok(ExchangeRate::new_unchecked(numerator, denominator))
+    }
+}
+
 impl Serial for ModuleReference {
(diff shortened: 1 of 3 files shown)


abizjak (approved):

abizjak:
Please state in the PR that is is moved from concordium-base.

concordium-contracts-common/src/impls.rs
```diff
@@ -163,6 +163,21 @@ impl Deserial for Amount {
...
+        out.write_u64(self.numerator())?;
+        out.write_u64(self.denominator())
+    }
+}
+
+impl Deserial for ExchangeRate {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        let numerator = source.read_u64()?;
+        let denominator = source.read_u64()?;
+        Ok(ExchangeRate::new_unchecked(numerator, denominator))
```
Bargsteen, line 177: Are we sure we want the unchecked version here? I know that we don't have `gcd` available as is, but perhaps we _should_ check that they are coprime?

Bargsteen: At least we could check that neither is `0`.

Bargsteen (commented):
Looks good, but I have one concern.

abizjak (approved):

concordium-contracts-common/src/impls.rs
abizjak: In practice this exchange rate will only be supplied by the node in serialized form. So we do not need to do many checks since the node will provide valid values.

Bargsteen: Yes, that does sound a bit surprising for the zero check. But ok, if we only ever get it from the node in practice, then it's fine.

Bargsteen (approved):

PR #72, opened 2023-02-28
Move `AccountBalance` from `concordium-std` and other small changes

Bargsteen:
## Purpose

A few smaller changes that I needed when working on https://github.com/Concordium/concordium-smart-contract-tools/pull/23.

## Changes

- Move `AccountBalance` from `concordium-std`.
- Add `to_owned` method to `EntrypointName` and `ContractName` types.
- Add `checked_sub` to `Amount` type.
- Add `from_bytes` and `empty` constructor methods to `OwnedParameter`.
- Add serialization instances for 4, 5, and 6 tuples.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-contracts-common/src/types.rs
+++ b/concordium-contracts-common/src/types.rs
@@ -355,6 +355,41 @@ impl ops::RemAssign<u64> for Amount {
     fn rem_assign(&mut self, other: u64) { *self = *self % other; }
 }
 
+/// The current public balances of an account.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct AccountBalance {
+    /// The total balance of the account. Note that part of this balance might
+    /// be staked and/or locked in scheduled transfers.
+    pub total:  Amount,
+    /// The current staked amount of the account. This amount is used for
+    /// staking.
+    pub staked: Amount,
+    /// The current amount locked in releases that resulted from transfers with
+    /// schedule. A locked amount can still be used for staking.
+    pub locked: Amount,
+}
+
+impl AccountBalance {
+    /// Construct a new account balance, ensuring that both the staked amount
+    /// and the locked amount is smaller than or equal to the total balance.
+    pub fn new(total: Amount, staked: Amount, locked: Amount) -> Option<Self> {
+        if total < staked || total < locked {
+            None
+        } else {
+            Some(Self {
+                total,
+                staked,
+                locked,
+            })
+        }
+    }
+
+    /// The current available balance of the account. This is the amount
+    /// an account currently have available for transfering and is not
+    /// staked or locked in releases by scheduled transfers.
+    pub fn available(&self) -> Amount { self.total - cmp::max(self.locked, self.staked) }
+}
+
 /// A reference to a smart contract module deployed on the chain.
 #[repr(transparent)]
 #[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
--- a/concordium-contracts-common/src/types.rs
+++ b/concordium-contracts-common/src/types.rs
@@ -225,11 +225,18 @@ impl Amount {
         }
     }
 
-    /// Checked addition. Adds another amount and return None if overflow
-    /// occurred
+    /// Checked addition. Adds another amount and returns None if overflow
+    /// occurred.
     #[inline(always)]
     pub fn checked_add(self, other: Amount) -> Option<Amount> {
         self.micro_ccd.checked_add(other.micro_ccd).map(Amount::from_micro_ccd)
+    }
+
+    /// Checked subtraction. Subtracts another amount and returns None if
+    /// underflow occurred.
+    #[inline(always)]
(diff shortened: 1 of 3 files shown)


concordium-contracts-common/src/traits.rs
```diff
@@ -196,7 +196,7 @@ impl Write for &mut [u8] {
     #[inline]
     fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Err> {
         let to_write = core::cmp::min(buf.len(), self.len());
-        let (overwrite, rest) = core::mem::replace(self, &mut []).split_at_mut(to_write);
```
abizjak, line 199: Well the thing is that I prefer using replace since it is more explicit. So why was this, unrelated, thing changed?

concordium-contracts-common/src/traits.rs
```diff
@@ -180,7 +180,7 @@ impl Write for Vec<u8> {
 
     #[inline]
     fn write(&mut self, buf: &[u8]) -> Result<usize, Self::Err> {
-        let _ = self.extend_from_slice(buf);
+        self.extend_from_slice(buf);
```
abizjak, line 183: Why is this unrelated thing changed?

concordium-contracts-common/src/types.rs
```diff
@@ -348,6 +356,41 @@ impl ops::RemAssign<u64> for Amount {
...
+            Some(Self {
+                total,
+                staked,
+                locked,
+            })
+        }
+    }
+
+    /// The current available balance of the account. This is the amount
+    /// an account currently have available for transfering and is not
```
abizjak, line 389:
```suggestion
    /// an account currently has available for transferring and is not
```

concordium-contracts-common/src/types.rs
```diff
@@ -859,6 +902,10 @@ impl<'a> ContractName<'a> {
     #[inline(always)]
     pub fn get_chain_name(self) -> &'a str { self.0 }
 
+    /// Convert a `ContractName` to its owned counterpart. This is an expensive
```
abizjak, line 905:
```suggestion
    /// Convert a [`ContractName`] to its owned counterpart. This is an expensive
```

concordium-contracts-common/src/types.rs
```diff
@@ -1115,6 +1162,10 @@ impl<'a> EntrypointName<'a> {
         Ok(Self(name))
     }
 
+    /// Convert a `EntrypointName` to its owned counterpart. This is an
```
abizjak, line 1165:
```suggestion
    /// Convert an [`EntrypointName`] to its owned counterpart. This is an
```

concordium-contracts-common/src/types.rs
```diff
@@ -1205,11 +1256,18 @@ impl From<Vec<u8>> for OwnedParameter {
...
     pub fn as_parameter(&self) -> Parameter { Parameter(self.0.as_ref()) }
 
-    /// Construct an `OwnedParameter` by serializing the input using its
+    /// Construct an [`Self`]` by serializing the input using its
     /// `Serial` instance.
     pub fn new<D: Serial>(data: &D) -> Self { Self(to_bytes(data)) }
+
+    /// Construct an [`Self`] from a vector of bytes.
+    pub fn from_bytes(bytes: Vec<u8>) -> Self { Self(bytes) }
+
```
abizjak, line 1268: I don't like this since we have a "from_bytes" method that deals with parsing/deserializing values. So that will cause confusion.

Moreover we would ideally be named with `_unchecked` since not every vector is a valid parameter.

concordium-contracts-common/src/traits.rs
Bargsteen: Clippy was complaining.

concordium-contracts-common/src/traits.rs
Bargsteen: Clippy was complaining.

abizjak: Well please use clippy that is in the CI, not a newer version. We don't want noise in PRs.

concordium-contracts-common/src/types.rs
```diff
@@ -1115,6 +1162,10 @@ impl<'a> EntrypointName<'a> {
         Ok(Self(name))
     }
 
+    /// Convert a `EntrypointName` to its owned counterpart. This is an
+    /// expensive operation that requires memory allocation.
```
limemloh, line 1166: Same comment as above related to calling this expensive

limemloh (approved):
LGTM

concordium-contracts-common/src/types.rs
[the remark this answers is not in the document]
abizjak: Well it is expensive when using it inside a contract.

Remarking that it allocates memory is not sufficient for many people.

I agree that when using this off-chain it is not expensive though.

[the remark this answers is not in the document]
Bargsteen: I'll leave it as is since it is expensive in the context of a contract.

abizjak (approved):

PR #75, opened 2023-03-08
Rename `new` to `from_serial` which ensures valid length

Bargsteen:
## Purpose

Replace `new` with `from_serial` which also checks the length.

## Changes

- Rename `new` to `from_serial` which ensures valid length
- Also add empty methods to both versions of parameter
- And implement Default for parameter

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-contracts-common/src/types.rs
+++ b/concordium-contracts-common/src/types.rs
@@ -1205,7 +1205,7 @@ impl<'a> Parameter<'a> {
 
     /// Construct an empty parameter.
     #[inline]
-    pub fn empty() -> Self { Self::default() }
+    pub fn empty() -> Self { Self(&[]) }
 }
 
 /// Parameter to the init function or entrypoint. Owned version.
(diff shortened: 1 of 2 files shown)


concordium-contracts-common/src/types.rs
```diff
@@ -1202,6 +1202,10 @@ impl<'a> Parameter<'a> {
     /// external means.
     #[inline]
     pub fn new_unchecked(bytes: &'a [u8]) -> Self { Self(bytes) }
+
+    /// Construct an empty parameter.
+    #[inline]
+    pub fn empty() -> Self { Self::default() }
```
abizjak, line 1208: I'd prefer if we constructed it directly here.

It would not be unreasonable to implement default in terms of empty. Which would lead to an infinite loop.

concordium-contracts-common/src/types.rs
```diff
@@ -1264,13 +1268,29 @@ impl OwnedParameter {
...
 
     /// Construct a parameter from a vector of bytes without checking that it
     /// fits the size limit. The caller is assumed to ensure this via
     /// external means.
     #[inline]
     pub fn new_unchecked(bytes: Vec<u8>) -> Self { Self(bytes) }
+
+    /// Construct an empty parameter.
+    #[inline]
+    pub fn empty() -> Self { Self::default() }
```
abizjak, line 1293: I'd prefer if we constructed it directly here.

It would not be unreasonable to implement default in terms of empty. Which would lead to an infinite loop.

abizjak (approved):

PR #84, opened 2023-05-02
Remove `sdk` feature, update version and edition, add helper function

Bargsteen:
## Purpose

A number of small changes needed for the integration testing library.

## Changes

- Add the method
- The newest version of `rust-decimal` requires rust 1.60, so the rust version is also updated.
- Update the Rust edition to `2021`
  - Used `cargo fix --edition` to put spaces in the macros.
- Update rust version in CI 
- Add contributor section in README which also explains the nightly version needed for `cargo fmt`
- Remove the SDK feature

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-contracts-common-derive/Cargo.toml
+++ b/concordium-contracts-common-derive/Cargo.toml
@@ -2,7 +2,8 @@
 name = "concordium-contracts-common-derive"
 authors = ["Concordium <developers@concordium.com>"]
 version = "1.0.1"
-edition = "2018"
+edition = "2021"
+rust-version = "1.60"
 license = "MPL-2.0"
 description = "Procedural macros to ease writing for smart contracts on the Concordium blockchain."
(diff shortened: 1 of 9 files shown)


abizjak (approved):

concordium-contracts-common-derive/Cargo.toml
```diff
@@ -2,16 +2,14 @@
 name = "concordium-contracts-common-derive"
 authors = ["Concordium <developers@concordium.com>"]
 version = "1.0.1"
```
abizjak, line 4: We should bump the version up so we do a new release of it.

Technically this is a breaking change so we need a new major.

concordium-contracts-common/Cargo.toml
```diff
@@ -2,7 +2,8 @@
 name = "concordium-contracts-common"
 version = "5.3.1"
```
abizjak, line 3: Bump version to 6 here as well.

PR #87, opened 2023-05-23
Improve serialization errors

soerenbf:
## Purpose

Adds trace information for errors happening when (de)serializing using contract schemas. I added a few unit tests, mostly to test the format of the trace errors for both `JsonError` and `ToJsonError` are as expected, however they are nowhere exhaustive (If we need more, let me know).

## Changes

- Adds `JsonError::TraceError`, which holds information about the path to the original `JsonError` produced when serializing from schema type
- Adds `ToJsonError` type, which is now the error type returned when deserializing to schema type.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-contracts-common/CHANGELOG.md
+++ b/concordium-contracts-common/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## Unreleased changes
 
 - Implement `serde::Serialize` and `serde::Deserialize` for `Duration` using `FromStr` and `Display` implementations, when feature `derive-serde` is enabled.
 
--- a/concordium-contracts-common/CHANGELOG.md
+++ b/concordium-contracts-common/CHANGELOG.md
@@ -1,8 +1,10 @@
 # Changelog
 
-## concordium-contracts-common 7.0.0 (2023-05-23)
+## Unreleased
 
 - Implement `serde::Serialize` and `serde::Deserialize` for `Duration` using `FromStr` and `Display` implementations, when feature `derive-serde` is enabled.
+
+### Breaking changes
 - Adds new error type `ToJsonError`, which is returned when deserializing to schema type fails.
 - Adds member `JsonError::TraceError` to `JsonError`, which adds trace information to the error produced when serializing from schema type fails.
 
--- a/concordium-contracts-common/CHANGELOG.md
+++ b/concordium-contracts-common/CHANGELOG.md
@@ -1,8 +1,10 @@
 # Changelog
 
-## Unreleased changes
+## concordium-contracts-common 7.0.0 (2023-05-23)
 
 - Implement `serde::Serialize` and `serde::Deserialize` for `Duration` using `FromStr` and `Display` implementations, when feature `derive-serde` is enabled.
+- Adds new error type `ToJsonError`, which is returned when deserializing to schema type fails.
+- Adds member `JsonError::TraceError` to `JsonError`, which adds trace information to the error produced when serializing from schema type fails.
 
 ## concordium-contracts-common 6.0.0 (2023-05-08)
 
--- a/concordium-contracts-common/src/schema_json.rs
+++ b/concordium-contracts-common/src/schema_json.rs
@@ -48,8 +48,8 @@ pub enum JsonError {
 }
 
 impl JsonError {
-    /// Wraps a [JsonError] in a [JsonError::TraceError], providing a trace to the origin of the
-    /// error.
+    /// Wraps a [JsonError] in a [JsonError::TraceError], providing a trace to
+    /// the origin of the error.
     fn add_trace(&self, trace: String) -> Self {
         JsonError::TraceError {
             trace,
@@ -57,8 +57,8 @@ impl JsonError {
         }
     }
 
-    /// Gets the underlying error of a [JsonError::TraceError]. For any other variant, this simply returns
-    /// the error itself.
+    /// Gets the underlying error of a [JsonError::TraceError]. For any other
+    /// variant, this simply returns the error itself.
     fn get_error(&self) -> Self {
         if let JsonError::TraceError {
             error,
@@ -292,13 +292,20 @@ fn write_bytes_from_json_schema_type<W: Write>(
             if let Value::Array(entries) = json {
                 let len = entries.len();
                 write_bytes_for_length_of_size(len, size_len, out)?;
+
+                let mut i = 0;
                 for entry in entries {
                     if let Value::Array(pair) = entry {
                         ensure!(pair.len() == 2, MapError("Expected key-value pair".to_string()));
-                        write_bytes_from_json_schema_type(key_ty, &pair[0], out)
-                            .map_err(|e| e.add_trace(format!("'{}' (key)", &pair[0])))?;
-                        write_bytes_from_json_schema_type(val_ty, &pair[1], out)
-                            .map_err(|e| e.add_trace(format!("'{}'", &pair[0])))?;
+                        let result: Result<(), JsonError> = {
+                            write_bytes_from_json_schema_type(key_ty, &pair[0], out)
+                                .map_err(|e| e.add_trace("0".to_string()))?;
+                            write_bytes_from_json_schema_type(val_ty, &pair[1], out)
+                                .map_err(|e| e.add_trace("1".to_string()))?;
+                            Ok(())
+                        };
+                        result.map_err(|e| e.add_trace(format!("{}", i)))?;
+                        i += 1;
                     } else {
                         return Err(WrongJsonType(
                             "Expected key value pairs as JSON arrays".to_string(),
@@ -964,9 +971,7 @@ impl Fields {
 }
 
 impl From<std::string::FromUtf8Error> for ParseError {
-    fn from(_: std::string::FromUtf8Error) -> Self {
-        ParseError::default()
-    }
+    fn from(_: std::string::FromUtf8Error) -> Self { ParseError::default() }
 }
 
 fn item_list_to_json<R: Read>(
--- a/concordium-contracts-common/src/schema_json.rs
+++ b/concordium-contracts-common/src/schema_json.rs
@@ -153,10 +153,14 @@ pub enum ToJsonError {
     #[error("Failed to format as JSON")]
     FormatError,
     /// Failed to deserialize data to type expected from schema.
-    #[error("Failed to deserialize {schema:?} from position {position} of bytes {data}")]
+    #[error(
+        "Failed to deserialize {schema:?} due to: {reason} - from position {position} of bytes \
+         {data}"
+    )]
     DeserialError {
         position: u32,
         schema:   Type,
+        reason:   String,
         data:     ToJsonErrorData,
     },
     /// Trace leading to the original [ToJsonError].
@@ -239,6 +243,10 @@ impl<'a> ToJsonError {
 }
 
 pub type ToJsonResult<A> = Result<A, ToJsonError>;
+
+#[derive(thiserror::Error, Debug, Clone)]
+#[error("{0}")]
+struct ParseErrorWithReason(String);
 
 // Serializes a value to the provided output buffer
 macro_rules! serial {
@@ -1344,6 +1352,7 @@ fn item_list_to_json<T: AsRef<[u8]>>(
     let len = deserial_length(source, size_len).map_err(|_| ToJsonError::DeserialError {
         data,
         position,
+        reason: "Could not deserialize length of list".into(),
         schema: schema.to_owned(),
     })?;
     let mut values = Vec::with_capacity(std::cmp::min(MAX_PREALLOCATED_CAPACITY, len));
@@ -1354,20 +1363,38 @@ fn item_list_to_json<T: AsRef<[u8]>>(
     Ok(values)
 }
 
-fn deserial_string<R: Read>(source: &mut R, size_len: SizeLength) -> ParseResult<String> {
-    let len = deserial_length(source, size_len)?;
+fn deserial_string<R: Read>(
+    source: &mut R,
+    size_len: SizeLength,
+) -> Result<String, ParseErrorWithReason> {
+    let len = deserial_length(source, size_len)
+        .map_err(|_| ParseErrorWithReason("Could not parse String length".into()))?;
     // we are doing this case analysis so that we have a fast path for safe,
     // most common, lengths, and a slower one longer ones.
     if len <= MAX_PREALLOCATED_CAPACITY {
         let mut bytes = vec![0u8; len];
-        source.read_exact(&mut bytes)?;
-        Ok(String::from_utf8(bytes)?)
+        source.read_exact(&mut bytes).map_err(|_| {
+            ParseErrorWithReason(format!(
+                "Parsed length {} for String value, but failed to read {} bytes from value",
+                len, len
+            ))
+        })?;
+        Ok(String::from_utf8(bytes)
+            .map_err(|_| ParseErrorWithReason("Failed to deserialize String from value".into()))?)
     } else {
         let mut bytes: Vec<u8> = Vec::with_capacity(MAX_PREALLOCATED_CAPACITY);
         let mut buf = [0u8; 64];
         let mut read = 0;
         while read < len {
-            let new = source.read(&mut buf)?;
+            let new = source.read(&mut buf).map_err(|_| {
+                let end = std::cmp::min(len - read, 64);
+                ParseErrorWithReason(format!(
+                    "Parsed length {} for String value, but failed to read bytes {}..{} from value",
+                    len,
+                    read,
+                    read + end
+                ))
+            })?;
             if new == 0 {
                 break;
             } else {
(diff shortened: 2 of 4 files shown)


concordium-contracts-common/CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
 # Changelog
 
-## Unreleased changes
+## concordium-contracts-common 7.0.0 (2023-05-23)
 
 - Implement `serde::Serialize` and `serde::Deserialize` for `Duration` using `FromStr` and `Display` implementations, when feature `derive-serde` is enabled.
+- Adds new error type `ToJsonError`, which is returned when deserializing to schema type fails.
+- Adds member `JsonError::TraceError` to `JsonError`, which adds trace information to the error produced when serializing from schema type fails.
```
abizjak, line 7: Breaking changes are missing in this list.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -40,8 +42,133 @@ pub enum JsonError {
...
+pub enum ToJsonError<'a> {
+    /// JSON formatter failed to represent value.
+    #[error("Failed to format as JSON")]
+    FormatError {},
+    /// Failed to deserialize data to type expected from schema.
+    #[error("Failed to deserialize {schema:?} from position {position} of bytes {data}")]
+    DeserialError {
+        position: u32,
+        schema:   &'a Type,
+        data:     ToJsonErrorData,
```
abizjak, line 125: It looks like in a number of cases you could provide some details here, for example "integer overflow".

Is there a reason for not adding an error message string here?

concordium-contracts-common/src/schema_json.rs
```diff
@@ -40,8 +42,133 @@ pub enum JsonError {
...
+    },
+    /// Trace leading to the original [ToJsonError].
+    #[error("{schema:?} -> {error}")]
```
abizjak, line 128: Isn't this going to be very verbose? Type is very large.

Bargsteen (commented):
Nice work 😊 
It would be nice, as Ales suggested, that we include the more precise reason why an individual element could not be converted. You can draw inspiration from [concordium client]( https://github.com/Concordium/concordium-client/blob/334e8d060875d9f486bfaa8c72f45d13931b024a/src/Concordium/Client/Types/Contract/Parameter.hs#LL61C4-L61C4).

concordium-contracts-common/src/schema_json.rs
```diff
@@ -40,8 +42,133 @@ pub enum JsonError {
...
+#[derive(Debug, Clone)]
+pub struct ToJsonErrorData(Vec<u8>);
+
+impl From<Vec<u8>> for ToJsonErrorData {
+    fn from(value: Vec<u8>) -> Self { Self(value) }
+}
+
+impl Display for ToJsonErrorData {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        self.0.iter().map(|b| write!(f, "{:02x?}", b)).collect()
```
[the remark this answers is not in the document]
soerenbf: It's a `try_for_each` instead now (clippy lead me on that path), which should be the same?

concordium-contracts-common/src/schema_json.rs
```diff
@@ -228,8 +357,12 @@ fn write_bytes_from_json_schema_type<W: Write>(
             if let Value::Array(values) = json {
                 let len = values.len();
                 write_bytes_for_length_of_size(len, size_len, out)?;
+
+                let mut i = 0;
                 for value in values {
```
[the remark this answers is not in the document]
soerenbf: yes, clippy suggested this as well 😄

concordium-contracts-common/src/schema_json.rs
```diff
@@ -40,8 +42,133 @@ pub enum JsonError {
...
+    pub fn print(&self, verbose: bool) -> String {
+        match self {
+            JsonError::TraceError {
+                field,
+                json,
+                error,
+            } if verbose => {
+                let formatted_json =
+                    serde_json::to_string_pretty(json).unwrap_or(format!("{}", json));
+                format!("{}\nIn {} of {}", error.print(verbose), field, formatted_json)
```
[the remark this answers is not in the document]
Bargsteen: I don't think we have a known bound here, so recursion should be avoided.
You can loop through the recursive structure here as well. Similar to what Ales suggested here: https://github.com/Concordium/concordium-contracts-common/pull/87#discussion_r1202057797.

concordium-contracts-common/src/schema_json.rs
soerenbf: Well.. yes. The verbose version is alright as it's on multiple lines, but the concise version (as in the one on this line) is not very nice to read. Maybe we just remove the type (`{schema:?}`) altogether from this and only print the error (which also points you to the precise position of the buffer that fails to deserialize into a concrete type).

Or do you have an alternative suggestion to what could be nice if it was to be included in f.x. a log file?

concordium-contracts-common/src/schema_json.rs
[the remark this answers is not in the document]
soerenbf: Tried to do something here. I don't think it's very pretty though, so any suggestions as to how to make it better is more than welcome!

concordium-contracts-common/src/schema_json.rs
```diff
@@ -962,178 +1322,251 @@ fn deserial_string<R: Read>(source: &mut R, size_len: SizeLength) -> ParseResult
...
-                let values = item_list_to_json(source, *size_len, |s| {
-                    let key = key_type.to_json(s)?;
-                    let value = value_type.to_json(s)?;
-                    Ok(Value::Array(vec![key, value]))
-                })?;
+                let values = item_list_to_json(
+                    source,
+                    *size_len,
+                    |s| {
+                        let key = key_type.to_json(s).map_err(|e| e.add_trace(position, self))?;
```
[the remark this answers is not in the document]
soerenbf: My thought was that the position in the `TraceError` marked the start of the point which is attempted to be deserialized as `Type::Map`, whereas the specific position that leads to the innermost error is included in the `DeserialError`, if that makes sense?

concordium-contracts-common/src/schema_json.rs
```diff
@@ -1,11 +1,13 @@
...
+use core::fmt::Display;
 use num_bigint::{BigInt, BigUint};
 use num_traits::Zero;
 use serde_json::Value;
 use std::convert::{TryFrom, TryInto};
 
+/// Represents errors occuring while serializing data from schema JSON format
 #[derive(Debug, thiserror::Error, Clone)]
-pub enum JsonError {
+pub enum JsonError<'a> {
```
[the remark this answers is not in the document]
soerenbf: Resolved this by having the error own the value instead (cloned the data). Tried to do this in a lazy fashion to limit this behaviour to happen only when needed

concordium-contracts-common/src/schema_json.rs
```diff
@@ -886,10 +1142,107 @@ mod tests {
...
+    }
 }
 
 impl Fields {
-    pub fn to_json<R: Read>(&self, source: &mut R) -> ParseResult<serde_json::Value> {
+    pub fn to_json<'a, T: AsRef<[u8]>>(
+        &'a self,
+        source: &mut Cursor<T>,
+    ) -> ToJsonResult<'a, serde_json::Value> {
         use serde_json::*;
```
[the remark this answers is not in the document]
soerenbf: Resolved with the removal of explicit lifetime annotation for errors altogether

concordium-contracts-common/src/schema_json.rs
abizjak: No, I can't really think of a good way to represent this in a single line.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -40,8 +42,133 @@ pub enum JsonError {
...
+            }
+            _ => format!("{}", self),
+        }
+    }
+}
+
+/// Wrapper around a list of bytes to represent data which failed to be
+/// deserialized into schema type.
+#[derive(Debug, Clone)]
+pub struct ToJsonErrorData(Vec<u8>);
```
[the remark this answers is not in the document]
abizjak: No, that covers it. Making the field `pub` would also do it. Unless there is a reason for not allowing its construction.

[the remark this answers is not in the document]
soerenbf: Ah, right 😅 it's now public!

concordium-contracts-common/src/schema_json.rs
```diff
@@ -953,192 +1472,347 @@ fn deserial_string<R: Read>(source: &mut R, size_len: SizeLength) -> ParseResult
...
             Type::Amount => {
-                let n = Amount::deserial(source)?;
+                let n = Amount::deserial(source)
+                    .map_err(|_| deserial_error("Could not parse Amount from value".into()))?;
                 Ok(Value::String(n.micro_ccd().to_string()))
             }
             Type::AccountAddress => {
-                let address = AccountAddress::deserial(source)?;
+                let address = AccountAddress::deserial(source).map_err(|_| {
+                    deserial_error("Could not parse AccountAddress from value".into())
```
abizjak, line 1576: And 32 bytes here.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -953,192 +1472,347 @@ fn deserial_string<R: Read>(source: &mut R, size_len: SizeLength) -> ParseResult
...
+                })?;
                 Ok(Value::String(address.to_string()))
             }
             Type::ContractAddress => {
-                let address = ContractAddress::deserial(source)?;
-                Ok(serde_json::to_value(address).map_err(|_| ParseError {})?)
+                let address = ContractAddress::deserial(source).map_err(|_| {
+                    deserial_error("Could not parse ContractAddress from value".into())
```
abizjak, line 1582: And here 16.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -953,192 +1472,347 @@ fn deserial_string<R: Read>(source: &mut R, size_len: SizeLength) -> ParseResult
...
+                        bytes.into_iter().collect()
+                    });
+
+                match (len, bytes) {
+                    (Ok(len), Ok(bytes)) => {
+                        let mut string = String::with_capacity(std::cmp::min(
+                            MAX_PREALLOCATED_CAPACITY,
+                            2 * len,
+                        ));
+                        bytes.into_iter().for_each(|b| string.push_str(&format!("{:02x?}", b)));
```
abizjak, line 1769: I am really not a fan of these effectful closures. I would strongly prefer to keep the for loop that we had.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -953,192 +1472,347 @@ fn deserial_string<R: Read>(source: &mut R, size_len: SizeLength) -> ParseResult
...
-                let mut string =
-                    String::with_capacity(std::cmp::min(MAX_PREALLOCATED_CAPACITY, 2 * len));
-                for _ in 0..len {
-                    let byte = source.read_u8()?;
-                    string.push_str(&format!("{:02x?}", byte));
+                let len = deserial_length(source, *size_len).map_err(|_| {
+                    ParseErrorWithReason("Could not parse ByteList length from value".into())
+                });
+                let bytes: core::result::Result<Vec<_>, _> =
+                    len.as_ref().map_err(|e| e.to_owned()).and_then(|len| {
```
abizjak, line 1746: Why `to_owned` instead of `clone`?

concordium-contracts-common/src/schema_json.rs
```diff
@@ -1,47 +1,301 @@
...
+            json: json.to_owned(),
+            error: Box::new(self),
+        }
+    }
+
+    /// Prints a formatted error message for variant. [JsonError::TraceError]
+    /// supports printing a verbose form including a more detailed
+    /// description of the error stack, which is returned if `verbose` is
+    /// set to true.
+    pub fn print(&self, verbose: bool) -> String { self.print_nested(verbose) }
```
Bargsteen, line 164: The method name here is also a bit misleading. If you change `print_nested`, then also change this one.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -1,47 +1,301 @@
 use crate::{constants::*, schema::*, *};
+use core::fmt::Display;
 use num_bigint::{BigInt, BigUint};
 use num_traits::Zero;
 use serde_json::Value;
 use std::convert::{TryFrom, TryInto};
 
+trait TraceError {
```
Bargsteen, line 8: Please document this trait.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -1,47 +1,301 @@
...
+    /// Gets the underlying error of a [ToJsonError::TraceError]. For any other
+    /// variant, this simply returns the error itself.
+    pub fn get_error(&self) -> &Self { self.get_innermost_error() }
+}
+
+pub type ToJsonResult<A> = Result<A, ToJsonError>;
+
+#[derive(thiserror::Error, Debug, Clone)]
+#[error("{0}")]
+struct ParseErrorWithReason(String);
```
Bargsteen, line 297: Docs, even though it is private ;)

concordium-contracts-common/src/schema_json.rs
```diff
@@ -756,6 +1031,117 @@ mod tests {
...
+        let schema = Type::AccountAddress;
+        let bytes =
+            schema.serial_value(&json!(format!("{}", &account))).expect("Serializing failed");
+
+        let expected = Vec::from(account_bytes);
+        assert_eq!(expected, bytes)
+    }
+
+    #[test]
+    fn test_serial_account_address_wrong_address_fails() {
```
Bargsteen, line 1047: Please add a brief comment to each of these, explaining what they are testing.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -886,10 +1272,114 @@ mod tests {
         let mut cursor = Cursor::new([128, 128, 128, 128, 128, 128, 128, 128, 128, 0b0111_1111]);
         deserial_bigint(&mut cursor, 9).expect_err("Deserialising should fail");
     }
+
+    #[test]
+    fn test_deserial_account_address() {
```
Bargsteen, line 1277: Also here. It might be redundant on this particular test case, but it will make the next person adding tests more likely to feel pressure to document his or her test case ;)

concordium-contracts-common/src/schema_json.rs
```diff
@@ -931,20 +1431,39 @@ fn item_list_to_json<R: Read>(
     Ok(values)
 }
 
-fn deserial_string<R: Read>(source: &mut R, size_len: SizeLength) -> ParseResult<String> {
-    let len = deserial_length(source, size_len)?;
+/// Deserialize a [String] of variable length from `source`.
```
Bargsteen, line 1434:
```suggestion
/// Deserialize a [`String`] of variable length from `source`.
```

concordium-contracts-common/src/schema_json.rs
```diff
@@ -931,20 +1431,39 @@ fn item_list_to_json<R: Read>(
...
         let mut read = 0;
         while read < len {
-            let new = source.read(&mut buf)?;
+            let new = source.read(&mut buf).map_err(|_| {
+                let end = std::cmp::min(len - read, 64);
+                ParseErrorWithReason(format!(
+                    "Parsed length {} for String value, but failed to read bytes {}..{} from value",
+                    len,
+                    read,
+                    read + end
```
Bargsteen, line 1464: This is very precise. I like it!

Bargsteen (approved):
Nice work! I just had some minor comments to add to Ales\.

concordium-contracts-common/src/schema_json.rs
soerenbf: No reason, if you prefer `clone` I can do that instead. Care to explain why you  prefer one over the other though?

abizjak: Clone is more direct and I think what you want here since you are going from `&A` to `A`.

`ToOwned` is more generic in that it allows you to change types, e.g., go from `&str` to `String`.

Of course there is a generic implementation of ToOwned for any Clone https://doc.rust-lang.org/src/alloc/borrow.rs.html#82 which makes these equivalent in this case, but I think it is better to be direct rather than generic.

soerenbf: Thanks for the elaborate response! 👍

concordium-contracts-common/src/schema_json.rs
```diff
@@ -1,47 +1,306 @@
...
+    /// including a more detailed description of the error stack, which is
+    /// returned if `verbose` is set to true.
+    pub fn print(&self, verbose: bool) -> String { self.display_nested(verbose) }
+
+    /// Gets the underlying error of a [`ToJsonError::TraceError`]. For any
+    /// other variant, this simply returns the error itself.
+    pub fn get_error(&self) -> &Self { self.get_innermost_error() }
+}
+
+/// Result with error type [`ToJsonError`]
```
abizjak, line 296: I think it would make sense to add an example or two of errors that can occcur. I think that would help explain the format pretty quickly. You can just take one of the nested examples from the test.

If we add that documentation here or to the method then it will be present in the rendered documentation on docs.rs which help make it understandable I think.

abizjak (approved):
Looks good.

But please add an example or two of how the rendered error looks like to exemplify how the path/indexing work with arrays and nested fields.

concordium-contracts-common/src/schema_json.rs
soerenbf: Good idea

soerenbf: Would it not make more sense to add this documentation to the actual error instead of the result, though?

soerenbf: I added examples for both `ToJsonError` and `JsonError` in each of their documentation blocks.

abizjak: > Would it not make more sense to add this documentation to the actual error instead of the result, though?

Sure. Whatever is most discoverable in the rendered documentation is fine with me.

abizjak: Actually the reason I suggested to add it to the type was that it is a type alias.

So in the siganture of the function it will just show ToJsonResult<A> and the error part will not be clickable directly.

So I thought it would not be easily discoverable.

soerenbf: Good point. I will remove the type alias instead.

PR #88, opened 2023-05-25
Conditionally derive Serial and Deserial for types with type parameters

limemloh:
## Purpose

Related to https://github.com/Concordium/concordium-rust-smart-contracts/issues/283.

## Changes

Add trait predicate for each type parameter when deriving `Serial` and `Deserial`.

## Checklist

      hard-to-understand areas.


concordium-contracts-common-derive/src/lib.rs
```diff
@@ -473,9 +488,27 @@ fn impl_serial(ast: &syn::DeriveInput) -> syn::Result<TokenStream> {
...
+            quote! (#type_param_ident: #root::Serial,)
+        })
+        .collect();
+
+    let where_clauses_tokens = if let Some(where_clauses) = where_clauses {
+        let predicates = &where_clauses.predicates;
+        quote!(#predicates, where_clause_serial)
+    } else {
+        where_clause_serial
+    };
```
abizjak, line 507: I think the implementation is fine as it is if we always want bounds for serial/deserial. 

But as we discussed I think it would be useful to be able to specify manual bounds like with `serde(bound)` (https://serde.rs/attr-bound.html )

Is there any problem with that?

limemloh:
I'll create another PR building on top of https://github.com/Concordium/concordium-contracts-common/pull/90

PR #91, opened 2023-06-01
Bound for type parameters

limemloh:
## Purpose

Related to https://github.com/Concordium/concordium-rust-smart-contracts/issues/283.
Tests are implemented in https://github.com/Concordium/concordium-rust-smart-contracts/pull/293.

## Changes

- Deriving `Serial`, `Deserial`, `DeserialWithState` and `SchemaType` now produces an implementation which adds a bound to each of the type parameters to implement the relevant trait.
  Note that `Serial` and `DeserialWithState` skips this bound for `state_parameter` when/if this is provided. This is not the behavior of `Deserial` and `SchemaType` since they are incompatible with `DeserialWithState` and therefore `state_parameter` is never present in these cases.
- Deriving `SchemaType` will now produce code even without the `build-schema` feature being enabled. This PR depends on it to avoid having to introduce separate types (like for `ContainerAttributes`) for when this feature is enabled.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -122,4 +122,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --all
+          args: --workspace
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -105,7 +105,7 @@ jobs:
           args: --manifest-path=concordium-contracts-common-derive/Cargo.toml --target=${{ matrix.target }} --no-default-features -- -D warnings
 
   test:
-    name: x86_64 tests on concordium-contracts-common
+    name: x86_64 tests
     runs-on: ubuntu-latest
     steps:
       - name: Checkout sources
@@ -122,4 +122,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --manifest-path=concordium-contracts-common/Cargo.toml
+          args: --all
--- a/concordium-contracts-common-derive/src/derive.rs
+++ b/concordium-contracts-common-derive/src/derive.rs
@@ -3,9 +3,7 @@
 use proc_macro::TokenStream;
 use proc_macro2::Span;
 use quote::ToTokens;
-#[cfg(feature = "build-schema")]
-use std::collections::HashMap;
-use std::{convert::TryFrom, ops::Neg};
+use std::{collections::HashMap, convert::TryFrom, ops::Neg};
 use syn::{spanned::Spanned, DataEnum, Ident, Meta};
 
 /// The prefix used in field attributes: `#[concordium(attr = "something")]`
@@ -15,7 +13,7 @@ const CONCORDIUM_ATTRIBUTE: &str = "concordium";
 const VALID_CONCORDIUM_FIELD_ATTRIBUTES: [&str; 3] = ["size_length", "ensure_ordered", "rename"];
 
 /// A list of valid concordium attributes
-const VALID_CONCORDIUM_ATTRIBUTES: [&str; 1] = ["state_parameter"];
+const VALID_CONCORDIUM_ATTRIBUTES: [&str; 2] = ["state_parameter", "bound"];
 
 fn get_root() -> proc_macro2::TokenStream { quote!(concordium_std) }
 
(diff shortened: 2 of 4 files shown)


concordium-contracts-common-derive/src/derive.rs
```diff
@@ -1148,3 +1459,74 @@ fn schema_type_fields(fields: &syn::Fields) -> syn::Result<proc_macro2::TokenStr
...
+        assert!(parsed.schema_type_bounds().is_none(), "Unexpected bound added for SchemaType");
+    }
+
+    /// Test parsing for bound attribute when using syntax for Serial explicit.
+    #[test]
+    fn test_parse_serial_explicit_bounds() {
+        let ast: syn::ItemStruct = syn::parse_quote! {
+            #[concordium(bound(serial  = "T: B"))]
+            struct MyStruct{}
+        };
```
abizjak, line 1506: This seems a bit of an odd test. I would not expect this to compile (I know this is not being tested here).

But it is an odd test case that the struct does not have a generic.

concordium-contracts-common-derive/src/lib.rs
```diff
@@ -44,33 +44,50 @@ pub fn deserial_derive(input: TokenStream) -> TokenStream {
...
+/// in source code. If the number of variants is less than or equal 256 then a
+/// single byte is used to encode it. Otherwise two bytes are used for the tag,
+/// encoded in little endian.
+///
+/// ## Generic type bounds
+///
+/// By default a trait bound is added on each generic type for implementing
+/// [`Serial`]. However, if this is not desirable, the bound can be put
+/// explicitly using the `bound` attribute on the type overriding the default
+/// behavior.
```
abizjak, line 66: Can you make it more explicit that this is added to the bounds present in the type declaration, i.e., if you have
```
#[concordium(T: Deserial)]
struct Foo<T: Deserial + Blah> {

}
```
then you get in total `T: Deserial + Blah + Deserial` as bounds (separated in where clause and type bounds).

I think a simple example is best really.

abizjak: In particular you cannot "get rid" of the type bounds using the `bound`.

concordium-contracts-common-derive/src/lib.rs
```diff
@@ -44,33 +44,50 @@ pub fn deserial_derive(input: TokenStream) -> TokenStream {
...
+/// ## Collections
 ///
 /// Collections (Vec, BTreeMap, BTreeSet) and strings (String, str) are by
 /// default serialized by prepending the number of elements as 4 bytes
 /// little-endian. If this is too much or too little, fields of the above types
 /// can be annotated with `size_length`.
 ///
 /// The value of this field is the number of bytes that will be used for
-/// encoding the number of elements. Supported values are 1, 2, 4, 8.
+/// encoding the number of elements. Supported values are `1,` `2,` `4,` `8`.
```
abizjak, line 85: Why not?
```suggestion
/// encoding the number of elements. Supported values are `1, 2, 4, 8`.
```

abizjak:
I was also thinking whether we should require that the bounds you supply only mention type variables that are actually declared, but it seems that is a lot of extra effort for something the compiler will already provide.

We could have a slightly better error message, but the information will still be provided so I don't think we need to do it.

.github/workflows/linter.yml
```diff
@@ -122,4 +122,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --manifest-path=concordium-contracts-common/Cargo.toml
+          args: --all
```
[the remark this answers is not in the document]
limemloh: Yes that was the intent

concordium-contracts-common-derive/src/derive.rs
```diff
@@ -157,15 +155,224 @@ fn find_state_parameter_attribute(
...
+            return Err(syn::Error::new(list.span(), "bound attribute cannot be empty"));
+        }
+        let mut deserial: Option<BoundAttributeValue> = None;
+        let mut serial: Option<BoundAttributeValue> = None;
+        let mut schema_type: Option<BoundAttributeValue> = None;
+
+        for item in items {
+            let syn::NestedMeta::Meta(nested_meta) = item else {
+                        return Err(syn::Error::new(item.span(), "bound attribute list can only contain name value pairs"));
+                    };
```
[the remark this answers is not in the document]
limemloh: It seems that Rustfmt is still working on supporting let-else statements https://github.com/rust-lang/rustfmt/issues/4914

I'll format this manually

concordium-contracts-common-derive/src/derive.rs
limemloh: I'll make it more like a real struct then

concordium-contracts-common-derive/src/lib.rs
limemloh: My intention was to emphasize that these are not string literals. The comma should not have been part of the monospacing

abizjak: Ok

abizjak (approved):

Bargsteen (approved):

PR #92, opened 2023-06-09
Support transparent attribute for deriving SchemaType

limemloh:
## Purpose

_Describe the purpose of the pull request, link to issue describing the problem, etc.

## Changes

_Describe the changes that were needed.

## Checklist

      hard-to-understand areas.


concordium-contracts-common-derive/src/derive.rs
```diff
@@ -1305,15 +1322,35 @@ pub fn schema_type_derive_worker(input: TokenStream) -> syn::Result<TokenStream>
...
+                if data.fields.len() != 1 {
+                    return Err(syn::Error::new(
+                        ast.span(),
+                        "'transparent' attribute can only be used on a struct with a single field",
+                    ));
+                }
+
+                // Safe to unwrap below since we already checked then length is one.
+                let field = data.fields.iter().next().unwrap();
+                schema_type_field_type(field)?
```
abizjak, line 1339:
```suggestion
                match data.fields.split_first() {
                    Some((field, rest)) if rest.is_empty() => {
                         schema_type_field_type(field)
                    _ => {
                    return Err(syn::Error::new(
                        ast.span(),
                        "'transparent' attribute can only be used on a struct with a single field",
                    ));
                }
```

concordium-contracts-common-derive/src/lib.rs
```diff
@@ -155,9 +155,75 @@ pub fn deserial_with_state_derive(input: TokenStream) -> TokenStream {
...
+/// ## Transparent
+///
+/// Deriving [`SchemaType`] for structs using the newtype design pattern exposes
+/// the wrapping struct which is often not desirable. The attribute
+/// `#[concordium(transparent)]` can be added above the struct which changes the
+/// implementation of [`SchemaType`] to schema type of the inner field.
+///
+/// The attribute can `#[concordium(transparent)]` attribute can only be used
+/// for structs with a single field, and the type of this field must implement
+/// `SchemaType`.
```
abizjak, line 216: There is a bit more complexity here. In a struct

```
struct Foo {
   #[concordium(size_length = 4)]
   inner: Vec<SomeType>
}
```

The field attribute is still used. This is important as well, but it is not obvious that this is a choice that was made.

So can you please also add an example with a vector or some other container and explain the semantics.

concordium-contracts-common-derive/src/derive.rs
```diff
@@ -264,6 +267,19 @@ impl TryFrom<&[syn::Attribute]> for ContainerAttributes {
                     },
                     Ok(bound) => bounds.push(bound),
                 }
+            } else if meta.path().is_ident("transparent") {
+                if let syn::Meta::Path(_) = meta {
+                    transparent = true
+                } else {
+                    let new_err = syn::Error::new(
+                        meta.span(),
+                        "transparent attribute cannot be a list or hold a value",
```
abizjak, line 276:
```suggestion
                        "'transparent' attribute cannot be a list or hold a value",
```

abizjak (approved):
Looks good, but please explain in the macro documentation the meaning of transparent in connection to field attributes.

concordium-contracts-common-derive/src/derive.rs
limemloh: I tried this just now, but `data.fields` does not have a method `split_first` and cannot become a slice easily.
https://docs.rs/syn/1.0.109/syn/enum.Fields.html

concordium-contracts-common-derive/src/lib.rs
limemloh: I have added a bit more to this

concordium-contracts-common-derive/src/derive.rs
abizjak: I think I looked at the wrong type then.

concordium-contracts-common-derive/src/derive.rs
```diff
@@ -1305,15 +1322,35 @@ pub fn schema_type_derive_worker(input: TokenStream) -> syn::Result<TokenStream>
...
-                concordium_std::schema::Type::Struct(#fields_tokens)
+            if container_attributes.transparent {
+                if data.fields.len() != 1 {
+                    return Err(syn::Error::new(
+                        ast.span(),
+                        "'transparent' attribute can only be used on a struct with a single field",
+                    ));
+                }
+
+                // Safe to unwrap below since we already checked then length is one.
```
Bargsteen, line 1337:
```suggestion
                // Safe to unwrap below since we already checked the length is one.
```

concordium-contracts-common-derive/src/lib.rs
```diff
@@ -155,9 +155,88 @@ pub fn deserial_with_state_derive(input: TokenStream) -> TokenStream {
...
 
-/// Derive the `SchemaType` trait for a type.
-/// If the feature `build-schema` is not enabled this is a no-op, i.e., it does
-/// not produce any code.
+/// Derive the [`SchemaType`] trait for a type matching the implementation when
+/// deriving [`Serial`].
+///
+/// Can be used for enums and structs.
+/// If the type is a struct all fields must implement the [`SchemaType`] trait.
+/// If the type is an enum then all fields of each of the enums must implement
```
Bargsteen, line 163:
```suggestion
/// If the type is an enum then all fields of each of the variants must implement
```

concordium-contracts-common-derive/src/lib.rs
```diff
@@ -155,9 +155,88 @@ pub fn deserial_with_state_derive(input: TokenStream) -> TokenStream {
     unwrap_or_report(derive::impl_deserial_with_state(&ast))
 }
 
-/// Derive the `SchemaType` trait for a type.
-/// If the feature `build-schema` is not enabled this is a no-op, i.e., it does
-/// not produce any code.
+/// Derive the [`SchemaType`] trait for a type matching the implementation when
+/// deriving [`Serial`].
```
Bargsteen, line 159: Is the fact that it matches `Serial` relevant? And is it always the case?
To me, the important part is just that I can use JSON.

concordium-contracts-common-derive/src/lib.rs
```diff
@@ -155,9 +155,88 @@ pub fn deserial_with_state_derive(input: TokenStream) -> TokenStream {
...
+/// If the type is a struct all fields must implement the [`SchemaType`] trait.
+/// If the type is an enum then all fields of each of the enums must implement
+/// the [`SchemaType`] trait.
+///
+/// ## Generic type bounds
+///
+/// By default a trait bound is added on each generic type for implementing
+/// [`SchemaType`]. However, if this is not desirable, the bound can be put
+/// explicitly using the `bound` attribute on the type overriding the default
+/// behavior by adding the provided bound for the implementation.
```
Bargsteen, line 171:
```suggestion
/// [`SchemaType`]. However, if this is not desirable, the default bound can be 
/// replaced by using the `bound` attribute on the type and providing the replacement.
```

concordium-contracts-common-derive/src/lib.rs
```diff
@@ -155,9 +155,88 @@ pub fn deserial_with_state_derive(input: TokenStream) -> TokenStream {
...
+/// ```
+///
+/// ## Transparent
+///
+/// Deriving [`SchemaType`] for structs using the newtype design pattern exposes
+/// the wrapping struct which is often not desirable. The attribute
+/// `#[concordium(transparent)]` can be added above the struct which changes the
+/// implementation of [`SchemaType`] to schema type of the field.
+///
+/// The attribute can `#[concordium(transparent)]` attribute can only be used
```
Bargsteen, line 214:
```suggestion
/// The `#[concordium(transparent)]` attribute can only be used
```

concordium-contracts-common-derive/src/lib.rs
```diff
@@ -155,9 +155,88 @@ pub fn deserial_with_state_derive(input: TokenStream) -> TokenStream {
...
+///
+/// ```ignore
+/// #[derive(SchemaType)]
+/// #[concordium(transparent)]
+/// struct Foo {
+///     bar: u32,
+/// }
+/// ```
+///
+/// ### Exmaple
```
Bargsteen, line 228:
```suggestion
/// ### Example
```

Bargsteen (approved):
LGTM, just minor corrections.

concordium-contracts-common-derive/src/lib.rs
limemloh: I find it relevant. It is a good point with the JSON, although when we have generated contract clients, this is not the only use case.

Issue #96, opened 2023-06-22
Implement `Display` Trait for `SchemaTypes`


DOBEN:
**Task description**

Adding the `Display` Trait and a function to display the schema in a human-readable manner. This is a prerequisite for adding a feature to `cargo-concordium` to decode and display schemas in a human-readable manner. We also intend to expose this functionality in the `concordium-web-sdk` so that front ends can decode/display schemas.

DOBEN:
close issue as completed

PR #97, opened 2023-06-22
Implement Display Trait for SchemaType

DOBEN:
## Purpose

addresses #96 

## Changes

Add `Display` trait for the Schema Type to display a template for the JSON to be used with the schema

### DIFF
--- a/concordium-contracts-common/CHANGELOG.md
+++ b/concordium-contracts-common/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+- Add `Display` trait to `SchemaType`
+
 ## concordium-contracts-common 7.0.0 (2023-06-16)
 
 - Implement `serde::Serialize` and `serde::Deserialize` for `Duration` using
--- a/concordium-contracts-common/CHANGELOG.md
+++ b/concordium-contracts-common/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased changes
 
-- Add `Display` trait to `SchemaType`
+- Add `Display` trait to `SchemaType` to display the `SchemaType` as a JSON template
 
 ## concordium-contracts-common 7.0.0 (2023-06-16)
 
--- a/concordium-contracts-common/src/schema_json.rs
+++ b/concordium-contracts-common/src/schema_json.rs
@@ -2,7 +2,7 @@ use crate::{constants::*, schema::*, *};
 use core::fmt::Display;
 use num_bigint::{BigInt, BigUint};
 use num_traits::Zero;
-use serde_json::Value;
+use serde_json::{json, Map, Value};
 use std::convert::{TryFrom, TryInto};
 
 /// Trait which includes implementations for unwrapping recursive error
@@ -963,6 +963,80 @@ impl Type {
         out: &mut W,
     ) -> Result<(), JsonError> {
         write_bytes_from_json_schema_type(schema, json, out)
+    }
+
+    // impl Display for Type {
+    //     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+    //         write!(f, "{}",
+    // serde_json::to_string_pretty(&self.to_json2()).unwrap())     }
+    // }
+    // List _ typ -> toJsonArray [AE.toJSON typ]
+    // Set _ typ -> toJsonArray [AE.toJSON typ]
+    // Map _ typK typV -> toJsonArray [toJsonArray [AE.toJSON typK, AE.toJSON typV]]
+    // Array _ typ -> toJsonArray [AE.toJSON typ]
+    // Struct fields -> AE.toJSON fields
+    // Enum variants -> AE.object ["Enum" .= (toJsonArray . map (\(k, v) ->
+    // AE.object [AE.fromText k .= v]) $ variants)] String _ -> AE.String "<String>"
+    // UInt128 -> AE.String "<UInt128>"
+    // Int128 -> AE.String "<Int128>"
+    // ContractName _ -> AE.object [ "contract" .= AE.String "<String>" ]
+    // ReceiveName _ -> AE.object [ "contract" .= AE.String "<String>", "func" .=
+    // AE.String "<String>" ] ULeb128 _ -> AE.String "<String with unsigned
+    // integer>" ILeb128 _ -> AE.String "<String with signed integer>"
+    // ByteList _ -> AE.String "<String with lowercase hex>"
+    // ByteArray _ -> AE.String "<String with lowercase hex>"
+    // TaggedEnum taggedVariants ->
+    //   let
+    //     variants = first (pack . show) <$> Map.toList taggedVariants
+    //   in
+    //     AE.object ["TaggedEnum" .= (toJsonArray . map (\(_k, (name, value)) ->
+    // AE.object [AE.fromText name .= value]) $ variants)] where toJsonArray =
+    // AE.Array . V.fromList
+
+    pub fn schema_to_json(&self) -> serde_json::Value {
+        match self {
+            Self::Unit => serde_json::Value::Array(Vec::new()),
+            Self::Bool => serde_json::Value::String("<Bool>".to_string()),
+            Self::U8 => serde_json::Value::String("<UInt8>".to_string()),
+            Self::U16 => serde_json::Value::String("<UInt16>".to_string()),
+            Self::U32 => serde_json::Value::String("<UInt32>".to_string()),
+            Self::U64 => serde_json::Value::String("<UInt64>".to_string()),
+            Self::U128 => serde_json::Value::String("<UInt128>".to_string()),
+            Self::I8 => serde_json::Value::String("<Int8>".to_string()),
+            Self::I16 => serde_json::Value::String("<Int16>".to_string()),
+            Self::I32 => serde_json::Value::String("<Int32>".to_string()),
+            Self::I64 => serde_json::Value::String("<Int64>".to_string()),
+            Self::I128 => serde_json::Value::String("<Int128>".to_string()),
+            Self::Amount => serde_json::Value::String("<Amount>".to_string()),
+            Self::AccountAddress => serde_json::Value::String("<AccountAddress>".to_string()),
+            Self::ContractAddress => {
+                let mut contract_address = Map::new();
+                contract_address.insert("index".to_string(), Type::U64.schema_to_json());
+                contract_address.insert("subindex".to_string(), Type::U64.schema_to_json());
+                serde_json::Value::Object(contract_address)
+            }
+            Self::Timestamp => serde_json::Value::String("<Timestamp>".to_string()),
+            Self::Duration => serde_json::Value::String("<Duration>".to_string()),
+            Self::Pair(a, b) => {
+                serde_json::Value::Array(vec![a.schema_to_json(), b.schema_to_json()])
+            }
+            Self::List(size, element) => serde_json::Value::Array(vec![element.schema_to_json()]),
+            Self::Set(size, element) => serde_json::Value::Array(vec![element.schema_to_json()]),
+            Self::Map(size, key, value) => serde_json::Value::Array(vec![
+                serde_json::Value::Array(vec![key.schema_to_json(), value.schema_to_json()]),
+            ]),
+            Self::Array(_, _) => serde_json::Value::String("<Bool>".to_string()),
+            Self::Struct(test) => serde_json::Value::String("<Bool>".to_string()),
+            Self::Enum(test) => serde_json::Value::String("<Bool>".to_string()),
+            Self::String(_) => serde_json::Value::String("<Bool>".to_string()),
+            Self::ContractName(_) => serde_json::Value::String("<Bool>".to_string()),
+            Self::ReceiveName(_) => serde_json::Value::String("<Bool>".to_string()),
+            Self::ULeb128(_) => serde_json::Value::String("<Bool>".to_string()),
+            Self::ILeb128(_) => serde_json::Value::String("<Bool>".to_string()),
+            Self::ByteList(_) => serde_json::Value::String("<Bool>".to_string()),
+            Self::ByteArray(_) => serde_json::Value::String("<Bool>".to_string()),
+            Self::TaggedEnum(_) => serde_json::Value::String("<Bool>".to_string()),
+        }
     }
 }
 
--- a/concordium-contracts-common/src/schema_json.rs
+++ b/concordium-contracts-common/src/schema_json.rs
@@ -930,6 +930,12 @@ fn write_bytes_from_json_schema_type<W: Write>(
     }
 }
 
+impl Display for Type {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "{}", serde_json::to_string_pretty(&self.schema_to_json()).unwrap())
+    }
+}
+
 impl Type {
     /// Serialize the given JSON value into the binary format represented by the
     /// schema. If the JSON value does not match the schema an error will be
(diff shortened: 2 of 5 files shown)


concordium-contracts-common/src/schema_json.rs
```diff
@@ -930,6 +930,41 @@ fn write_bytes_from_json_schema_type<W: Write>(
...
+                let mut map = Map::new();
+
+                for (name, field) in vec.iter() {
+                    map.insert(name.to_string(), field.schema_type_to_json());
+                }
+
+                serde_json::Value::Object(map)
+            }
+            Fields::Unnamed(vec) => {
+                let mut new_vector: Vec<Value> = vec![];
```
Bargsteen, line 946:
```suggestion
                let mut new_vector = Vec::new();
```
I think it is slightly more idiomatic to use `Vec::new()` instead of the macro for empty vectors. And can't the compiler infer the type?

concordium-contracts-common/src/schema_json.rs
```diff
@@ -930,6 +930,41 @@ fn write_bytes_from_json_schema_type<W: Write>(
...
+                for element in vec.iter() {
+                    new_vector.push(element.schema_type_to_json())
+                }
+
+                serde_json::Value::Array(new_vector)
+            }
+            Fields::None => {
+                let mut map = Map::new();
+                map.insert("None".to_string(), serde_json::Value::Array(Vec::new()));
+                serde_json::Value::Object(map)
```
Bargsteen, line 956: This should just be an empty vector. 
So that it will output:
`"MyEnumVariant" : []`

See https://github.com/Concordium/concordium-client/blob/30db64c1e3f4ba8d05c41080e52af47bcef909e3/src/Concordium/Client/Types/Contract/Schema.hs#L424

concordium-contracts-common/src/schema_json.rs
```diff
@@ -930,6 +930,41 @@ fn write_bytes_from_json_schema_type<W: Write>(
     }
 }
 
+impl Fields {
+    pub fn field_to_json(&self) -> serde_json::Value {
```
Bargsteen, line 934: Please document this method and the other one as well.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -930,6 +930,41 @@ fn write_bytes_from_json_schema_type<W: Write>(
...
+            Fields::None => {
+                let mut map = Map::new();
+                map.insert("None".to_string(), serde_json::Value::Array(Vec::new()));
+                serde_json::Value::Object(map)
+            }
+        }
+    }
+}
+
+impl Display for Type {
```
Bargsteen, line 962: Document that it displays it as JSON.

concordium-contracts-common/src/schema_json.rs
Bargsteen: Also mention that it should match the output of Concordium Client.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -964,6 +999,109 @@ impl Type {
     ) -> Result<(), JsonError> {
         write_bytes_from_json_schema_type(schema, json, out)
     }
+
+    pub fn schema_type_to_json(&self) -> serde_json::Value {
```
Bargsteen, line 1003: It would be good with a few tests, at least for some of the more complex types.
I am not sure what would be best, asserting on the strings, or just on the JSON types.

Bargsteen (changes requested):
Nice work, just needs a bit of polish ;)

concordium-contracts-common/src/schema_json.rs
abizjak: I don't think it's specified the order in which JSON fields are output in objects, so those tests will be fragile if we compare the output strings. We should compare `Value`s.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -964,6 +999,109 @@ impl Type {
...
+            }
+            Self::Amount => serde_json::Value::String("<Amount>".to_string()),
+            Self::AccountAddress => serde_json::Value::String("<AccountAddress>".to_string()),
+            Self::ContractAddress => {
+                let mut contract_address = Map::new();
+                contract_address.insert("index".to_string(), Type::U64.schema_type_to_json());
+                contract_address.insert("subindex".to_string(), Type::U64.schema_type_to_json());
+                serde_json::Value::Object(contract_address)
+            }
+            Self::Timestamp => serde_json::Value::String("<Timestamp>".to_string()),
```
limemloh, line 1067: An example for the date format would be nice for the user. Like `"2023-12-01T12:00:00Z"`.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -964,6 +999,109 @@ impl Type {
...
+            Self::Duration => serde_json::Value::String("<Duration>".to_string()),
```
limemloh, line 1068: An example of the format would be nice. Like `10d 1h 42s`

concordium-contracts-common/src/schema_json.rs
```diff
@@ -964,6 +999,109 @@ impl Type {
...
+                outer_map.insert("Enum".to_string(), serde_json::Value::Array(vector));
+
+                serde_json::Value::Object(outer_map)
+            }
+            Self::Struct(field) => field.field_to_json(),
+            Self::ByteList(_) => {
+                serde_json::Value::String("<String with lowercase hex>".to_string())
+            }
+            Self::ByteArray(_) => {
+                serde_json::Value::String("<String with lowercase hex>".to_string())
```
limemloh, line 1038: The expected length could be provided here as well

concordium-contracts-common/src/schema_json.rs
```diff
@@ -964,6 +999,109 @@ impl Type {
...
+                contract_address.insert("subindex".to_string(), Type::U64.schema_type_to_json());
+                serde_json::Value::Object(contract_address)
+            }
+            Self::Timestamp => serde_json::Value::String("<Timestamp>".to_string()),
+            Self::Duration => serde_json::Value::String("<Duration>".to_string()),
+            Self::Pair(a, b) => {
+                serde_json::Value::Array(vec![a.schema_type_to_json(), b.schema_type_to_json()])
+            }
+            Self::List(_, element) => serde_json::Value::Array(vec![element.schema_type_to_json()]),
+            Self::Array(_, element) => {
```
limemloh, line 1073: Could we provide the expected length? maybe generate an array of this size?

concordium-contracts-common/src/schema_json.rs
```diff
@@ -964,6 +999,109 @@ impl Type {
...
+            }
+            Self::ReceiveName(_) => {
+                let mut receive_name = Map::new();
+                receive_name.insert(
+                    "contract".to_string(),
+                    serde_json::Value::String("<String>".to_string()),
+                );
+                receive_name
+                    .insert("func".to_string(), serde_json::Value::String("<String>".to_string()));
+                serde_json::Value::Object(receive_name)
```
limemloh, line 1101: Have you considered using the [`json!`](https://docs.rs/serde_json/latest/serde_json/macro.json.html) macro?
I think it could improve the readability of the code quite a bit.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -964,6 +999,109 @@ impl Type {
...
+            }
+            Self::Struct(field) => field.field_to_json(),
+            Self::ByteList(_) => {
+                serde_json::Value::String("<String with lowercase hex>".to_string())
+            }
+            Self::ByteArray(_) => {
+                serde_json::Value::String("<String with lowercase hex>".to_string())
+            }
+            Self::String(_) => serde_json::Value::String("<String>".to_string()),
+            Self::Unit => serde_json::Value::Array(Vec::new()),
```
limemloh, line 1041: To me `null` makes more sense here

concordium-contracts-common/src/schema_json.rs
limemloh, line 934: I think the name will be confusing if we later have a JSON format for the schema itself, so maybe add that this is a JSON template of the schema.
Also, since it is a method on Fields, I don't see the need to include it in the name.
```suggestion
    pub fn to_json_template(&self) -> serde_json::Value {
```

concordium-contracts-common/src/schema_json.rs
limemloh, line 1003: Just like one of my comments above, I think we should emphasize this is a JSON template and not the schema itself as JSON.
```suggestion
    pub fn to_json_template(&self) -> serde_json::Value {
```

concordium-contracts-common/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+- Add `Display` trait to `SchemaType`
```
limemloh, line 5: I think it is important to emphasize that this is displaying the schema type as a JSON template

limemloh:
Sorry for being pedantic, but this is not displaying the Schema in a human readable format, this is displaying a template for the JSON to use with the schema. If this is not the goal, then we need to encode all the information of the schema in the JSON.

concordium-contracts-common/src/schema_json.rs
Bargsteen: Yes, that would be nice. It is a bit different than what the client does, but it could be updated as well.
I still think the brackets make sense to indicate that it is a placeholder. Perhaps `"<Timestamp ("2000-12-24T12:00:00Z")>"`?

concordium-contracts-common/src/schema_json.rs
Bargsteen: Yes, see my comment for the timestamp.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -964,6 +999,109 @@ impl Type {
...
+            Self::I32 => serde_json::Value::String("<Int32>".to_string()),
+            Self::I64 => serde_json::Value::String("<Int64>".to_string()),
+            Self::I128 => serde_json::Value::String("<Int128>".to_string()),
+            Self::ILeb128(_) => {
+                serde_json::Value::String("<String with signed integer>".to_string())
+            }
+            Self::ULeb128(_) => {
+                serde_json::Value::String("<String with unsigned integer>".to_string())
+            }
+            Self::Amount => serde_json::Value::String("<Amount>".to_string()),
```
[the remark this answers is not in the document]
Bargsteen: Yep. `"<Amount in microCCD>"` (I assume it is in microCCD).

concordium-contracts-common/src/schema_json.rs
Bargsteen: But that is not how we encode it everywhere else.

DOBEN: I kept it as it is because that is what `concordium-client` does.

concordium-contracts-common/src/schema_json.rs
DOBEN: It is displayed as follows now:
```
<Timestamp (e.g. `2000-01-01T12:00:00Z`)>
```

concordium-contracts-common/src/schema_json.rs
DOBEN: It is displayed as follows now:
```
"<Duration (e.g. `10d 1h 42s`)>"
```

concordium-contracts-common/src/schema_json.rs
```diff
@@ -930,6 +930,41 @@ fn write_bytes_from_json_schema_type<W: Write>(
     }
 }
 
+/// Displays a template of the JSON to be used for the Fields.
+/// It should match the output of `concordium-client`.
+impl Fields {
+    pub fn to_json_template(&self) -> serde_json::Value {
```
Bargsteen, line 936:
```suggestion
impl Fields {
    /// Displays a template of the JSON to be used for the Fields.
    /// It should match the output of `concordium-client`.
    pub fn to_json_template(&self) -> serde_json::Value {
```

concordium-contracts-common/src/schema_json.rs
```diff
@@ -1087,6 +1223,140 @@ mod tests {
...
+    /// Tests schema template display in JSON of an Enum
+    #[test]
+    fn test_schema_template_enum() {
+        let schema = Type::Enum(Vec::from([
+            (String::from("Accounts"), Fields::Unnamed(Vec::from([Type::AccountAddress]))),
+            (String::from("Accounts"), Fields::Unnamed(Vec::from([Type::AccountAddress]))),
+        ]));
+        assert_eq!(
+            json!({"Enum": [{"Accounts": ["<AccountAddress>"]},{"Accounts": ["<AccountAddress>"]}]}),
+            schema.to_json_template()
```
Bargsteen, line 1235: This seems odd. Both variants have the same name. Perhaps you meant to display an `Address` with `AddressAccount` / `AddressContract` variants?

Bargsteen (approved):
Nice!

concordium-contracts-common/src/schema_json.rs
```diff
@@ -964,6 +1001,107 @@ impl Type {
...
+            Self::Amount => json!("<Amount in microCCD>".to_string()),
+            Self::AccountAddress => json!("<AccountAddress>".to_string()),
+            Self::ContractAddress => {
+                let mut contract_address = Map::new();
+                contract_address.insert("index".to_string(), Type::U64.to_json_template());
+                contract_address.insert("subindex".to_string(), Type::U64.to_json_template());
+                json!(contract_address)
+            }
+            Self::Timestamp => {
+                json!("<Timestamp (e.g. `2000-01-01T12:00:00Z`)>".to_string())
```
abizjak, line 1072:
```suggestion
                "<Timestamp (e.g. `2000-01-01T12:00:00Z`)>".into()
```

concordium-contracts-common/src/schema_json.rs
```diff
@@ -964,6 +1001,107 @@ impl Type {
...
+            Self::List(_, element) => json!(vec![element.to_json_template()]),
+            Self::Array(size, element) => {
+                let mut vec = Vec::new();
+                for _i in 0..*size {
+                    vec.push(element.to_json_template())
+                }
+                json!(vec)
+            }
+            Self::Set(_, element) => {
+                json!(vec![element.to_json_template()])
```
abizjak, line 1087:
```suggestion
                vec![element.to_json_template()].into()
```

concordium-contracts-common/src/schema_json.rs
```diff
@@ -964,6 +1001,107 @@ impl Type {
...
+            }
+            Self::Map(_, key, value) => {
+                json!(vec![json!(vec![key.to_json_template(), value.to_json_template(),])])
```
abizjak, line 1090:
```suggestion
                vec![json!(vec![key.to_json_template(), value.to_json_template(),])].into()
```

concordium-contracts-common/src/schema_json.rs
```diff
@@ -964,6 +1001,107 @@ impl Type {
...
+            }
+            Self::ContractName(_) => {
+                let mut contract_name = Map::new();
+                contract_name.insert("contract".to_string(), json!("<String>".to_string()));
```
abizjak, line 1094:
```suggestion
                contract_name.insert("contract".to_string(), "<String>".into());
```

concordium-contracts-common/src/schema_json.rs
```diff
@@ -964,6 +1001,107 @@ impl Type {
...
+                json!(contract_name)
+            }
+            Self::ReceiveName(_) => {
+                let mut receive_name = Map::new();
+                receive_name.insert("contract".to_string(), json!("<String>".to_string()));
+                receive_name.insert("func".to_string(), json!("<String>".to_string()));
```
abizjak, line 1100:
```suggestion
                receive_name.insert("func".to_string(), "<String>".into());
```

concordium-contracts-common/src/schema_json.rs
```diff
@@ -1087,6 +1225,140 @@ mod tests {
 
     use super::*;
```
abizjak, line 1227:
```suggestion
    use super::*;
    use std::collections::BTreeMap;

```

concordium-contracts-common/src/schema_json.rs
```diff
@@ -2,7 +2,9 @@ use crate::{constants::*, schema::*, *};
 use core::fmt::Display;
 use num_bigint::{BigInt, BigUint};
 use num_traits::Zero;
-use serde_json::Value;
+use serde_json::{json, Map, Value};
+#[allow(unused_imports)] // BTreeMap is used in the unit tests.
+use std::collections::BTreeMap;
```
abizjak, line 7:
```suggestion
```
Please just import it in tests below.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -930,6 +932,41 @@ fn write_bytes_from_json_schema_type<W: Write>(
     }
 }
 
+impl Fields {
+    /// Displays a template of the JSON to be used for the Fields.
+    /// It should match the output of `concordium-client`.
```
abizjak, line 937: That's not really a requirement. The requirement is that it has to be as much as possible to copy the template and edit it to provide the actual parameters.

The comment about concordium-client does not really tell the caller of this function very much.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -930,6 +932,41 @@ fn write_bytes_from_json_schema_type<W: Write>(
...
+
+                json!(new_vector)
+            }
+            Fields::None => serde_json::Value::Array(Vec::new()),
+        }
+    }
+}
+
+/// Displays a pretty-printed JSON-template of the schema.
+/// It should match the output of `concordium-client`.
```
abizjak, line 963: Same as above.

abizjak (approved):
Looks good.

limemloh (approved):
LGTM

PR #99, opened 2023-06-27
Display function for schema template

DOBEN:
## Purpose

addresses https://github.com/Concordium/concordium-contracts-common/issues/96

## Changes

Add Display trait for the VersionedModuleSchema to display a template for the JSON to be used with the schema

### DIFF
--- a/concordium-contracts-common/src/schema_json.rs
+++ b/concordium-contracts-common/src/schema_json.rs
@@ -931,7 +931,7 @@ fn write_bytes_from_json_schema_type<W: Write>(
 }
 
 impl Fields {
-    /// Displays a template of the JSON to be used for the Fields.
+    /// Displays a template of the JSON to be used for the [`Fields`].
     pub fn to_json_template(&self) -> serde_json::Value {
         match self {
             Fields::Named(vec) => {
@@ -956,16 +956,15 @@ impl Fields {
     }
 }
 
-/// Displays a pretty-printed JSON-template of the schema.
+/// Displays a pretty-printed JSON-template of the [`SchemaType`].
 impl Display for Type {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         write!(f, "{}", serde_json::to_string_pretty(&self.to_json_template()).unwrap())
     }
 }
 
-/// Displays the type value indented.
-/// It should match the output of `concordium-client`.
-fn display_type_schema_indented(
+/// Displays the json template of the type indented.
+fn display_json_template_indented(
     mut out: String,
     type_schema: &Type,
     type_schema_name: &str,
@@ -982,8 +981,7 @@ fn display_type_schema_indented(
     out
 }
 
-/// Displays a pretty-printed JSON-template of the schema.
-/// It should match the output of `concordium-client`.
+/// Displays a pretty-printed template of the [`VersionedModuleSchema`].
 impl Display for VersionedModuleSchema {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         let mut out = String::new();
@@ -1036,10 +1034,11 @@ impl Display for VersionedModuleSchema {
                         out = format!("{}{:>2}Init:\n", out, "");
 
                         if let Some(type_schema) = schema.parameter() {
-                            out = display_type_schema_indented(out, type_schema, "Parameter", 4)
+                            out = display_json_template_indented(out, type_schema, "Parameter", 4)
                         }
                         if let Some(type_schema) = schema.return_value() {
-                            out = display_type_schema_indented(out, type_schema, "Return value", 4)
+                            out =
+                                display_json_template_indented(out, type_schema, "Return value", 4)
                         }
                     }
 
@@ -1053,10 +1052,11 @@ impl Display for VersionedModuleSchema {
                         out = format!("{}{:>4}- {:?}\n", out, "", function_name);
 
                         if let Some(type_schema) = schema.parameter() {
-                            out = display_type_schema_indented(out, type_schema, "Parameter", 6)
+                            out = display_json_template_indented(out, type_schema, "Parameter", 6)
                         }
                         if let Some(type_schema) = schema.return_value() {
-                            out = display_type_schema_indented(out, type_schema, "Return value", 6)
+                            out =
+                                display_json_template_indented(out, type_schema, "Return value", 6)
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


concordium-contracts-common/src/schema_json.rs
```diff
@@ -956,13 +956,234 @@ impl Fields {
     }
 }
 
-/// Displays a pretty-printed JSON-template of the schema.
+/// Displays a pretty-printed JSON-template of the [`SchemaType`].
```
Bargsteen, line 959:
```suggestion
/// Displays a pretty-printed JSON-template of the [`Type`].
```
`SchemaType` is the name of the derive macro :blush:

concordium-contracts-common/src/schema_json.rs
```diff
@@ -956,13 +956,234 @@ impl Fields {
...
 }
 
+/// Displays the json template of the type indented.
+fn display_json_template_indented(
+    mut out: String,
+    type_schema: &Type,
+    type_schema_name: &str,
+    indent: usize,
+) -> String {
+    out = format!("{}{:>3$}{}:\n", out, "", type_schema_name, indent);
```
Bargsteen, line 973: Could you document what the `:>3$` does exactly?

concordium-contracts-common/src/schema_json.rs
```diff
@@ -956,13 +956,234 @@ impl Fields {
...
-/// Displays a pretty-printed JSON-template of the schema.
+/// Displays a pretty-printed JSON-template of the [`SchemaType`].
 impl Display for Type {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         write!(f, "{}", serde_json::to_string_pretty(&self.to_json_template()).unwrap())
     }
 }
 
+/// Displays the json template of the type indented.
+fn display_json_template_indented(
```
Bargsteen, line 967: Could you document this method a bit more?
How will it be formatted and what do the parameters mean exactly? I see that `2` is always added to `indent` in the second call of `format` which seems a bit odd if you give a large `indent` value.

concordium-contracts-common/src/schema_json.rs
```diff
@@ -956,13 +956,234 @@ impl Fields {
...
+        out,
+        "",
+        type_schema.to_string().replace('\n', "\n        "),
+        indent + 2
+    );
+    out
+}
+
+/// Displays a pretty-printed template of the [`VersionedModuleSchema`].
+impl Display for VersionedModuleSchema {
```
Bargsteen, line 985: Could you add an example output to the docs, so we can see what the code is supposed to do?

concordium-contracts-common/src/schema_json.rs
```diff
@@ -1048,10 +1269,10 @@ impl Type {
             Self::I64 => "<Int64>".into(),
             Self::I128 => "<Int128>".into(),
             Self::ILeb128(size) => {
-                format!("String of size at most {size} containing a signed integer.").into()
+                format!("<String of size at most {size} containing a signed integer.>").into()
```
limemloh, line 1272: The size here is a bit more complex and a bit hard to explain, but it is the max number of bytes for the leb128 encoding, the user types in the integer (wrapped in a string) and not the encoding, so not sure we can communicate the size in a good way

concordium-contracts-common/src/schema_json.rs
```diff
@@ -1032,7 +1253,7 @@ impl Type {
             Self::Struct(field) => field.to_json_template(),
             Self::ByteList(_) => "<String with lowercase hex>".into(),
             Self::ByteArray(size) => {
-                format!("String of size {size} containing lowercase hex characters.").into()
+                format!("<String of size {size} containing lowercase hex characters.>").into()
```
limemloh, line 1256: Technically, the string itself will be twice the size because it uses two characters per byte

limemloh (approved):

concordium-contracts-common/src/schema_json.rs
DOBEN: Explanation:

The value `""` is `right-aligned in width columns`. Meaning there is just `tabs/whitespace` created here.
https://doc.rust-lang.org/std/fmt/#fillalignment

The amount of `tabs/whitespace` created is defined in the third argument (`3$`), meaning the `indent` defines the amount of `tabs/whitespace` created.

Does the explanation clarify it? I would keep it as it is if ok since people could look up the meaning of the symbols e.g. here  `https://doc.rust-lang.org/std/fmt/#fillalignment`.

concordium-contracts-common/src/schema_json.rs
DOBEN: Added the following:
```
/// Displays the json template of the type indented.
/// The `type_schema_name` is indented by `indent` and the associated `type_schema` is indented by `indent + 2`.
/// The `type_schema` is one level deeper than the `type_schema_name` and therefore indented by `+ 2`.
/// '\n' is replaced by "\n        " because the `VersionedModuleSchema` template has several higher level outputs (e.g. `contractNames`, `functionNames`, ...).
```

concordium-contracts-common/src/schema_json.rs
DOBEN: I changed it to `2*size` which seems more intuitive if people check the size of the `string`.

Bargsteen (commented):
Ah, I forgot to submit this yesterday. Sorry about that.

Bargsteen (approved):

concordium-contracts-common/src/schema_json.rs
DOBEN: Added some additional explanation about the `:>3$` syntax and also added the above link in a comment.

limemloh (approved):
LGTM

concordium-contracts-common/src/schema_json.rs
```diff
@@ -956,13 +956,305 @@ impl Fields {
...
+///
+/// assert_eq!(display, format!("{}", schema));
+/// ```
+impl Display for VersionedModuleSchema {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        let mut out = String::new();
+        match self {
+            VersionedModuleSchema::V0(module_v0) => {
+                for (contract_name, contract_schema) in module_v0.contracts.iter() {
+                    out = format!("Contract: {:>30}\n", contract_name);
```
abizjak, line 1056: Why do we put so much whitespace here?

concordium-contracts-common/src/schema_json.rs
```diff
@@ -956,13 +956,305 @@ impl Fields {
...
+                    // State
+                    if let Some(type_schema) = &contract_schema.state {
+                        out = format!("{}{:>2}State:\n", out, "");
+                        out = format!(
+                            "{}{:>4}{}\n",
+                            out,
+                            "",
+                            type_schema.to_string().replace('\n', "\n    ")
```
abizjak, line 1064: I usually don't like this sort of formatting, and would prefer if we could use a crate like https://crates.io/crates/indenter for indenting data, which is also more efficient than converting to string and then replacing things.

But with the work that is already done, changing to using that is going to be quite a bit of effort, so if you've sufficiently tested this let's just go ahead and merge.

abizjak (approved):

concordium-contracts-common/src/schema_json.rs
DOBEN: I reduced it to:
```
out = format!("Contract: {:>11}\n", contract_name);
```

concordium-contracts-common/src/schema_json.rs
DOBEN: Good point. I will explore that crate and try to use it next time. For now, I left the current way of formatting the string.

PR #100, opened 2023-06-29
Adds function to parse VersionedModuleSchema from base64

soerenbf:
## Purpose

Needed this for the custom indexer epic, thought it could be useful elsewhere as well.

## Changes

- Add `from_base64_str` to `VersionedModuleSchema`

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-contracts-common/src/schema.rs
+++ b/concordium-contracts-common/src/schema.rs
@@ -8,6 +8,7 @@ use crate::{impls::*, traits::*, types::*};
 use alloc::boxed::Box;
 #[cfg(not(feature = "std"))]
 use alloc::{collections, string::String, vec::Vec};
+use base64::{engine::general_purpose, Engine};
 use collections::{BTreeMap, BTreeSet};
 #[cfg(not(feature = "std"))]
 use core::{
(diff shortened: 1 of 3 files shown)


concordium-contracts-common/src/schema.rs
```diff
@@ -8,6 +8,8 @@ use crate::{impls::*, traits::*, types::*};
 use alloc::boxed::Box;
 #[cfg(not(feature = "std"))]
 use alloc::{collections, string::String, vec::Vec};
+#[allow(unused_imports)]
```
abizjak, line 11: We don't do that here :) 

Please only import it in the context that it is needed in, so that allow(unused_imports) is not necerssary.

abizjak: Around line 1016 in the inner module.

concordium-contracts-common/src/schema.rs
```diff
@@ -1121,6 +1123,16 @@ mod impls {
             Ok(versioned_module_schema)
         }
 
+        /// Get a versioned module schema from a base64 string.
+        pub fn from_base64_str(s: &str) -> Result<Self, VersionedSchemaError> {
+            let bytes: Vec<u8> = general_purpose::STANDARD_NO_PAD
```
abizjak, line 1128: In principle the input should always be padded, but using NO_PAD is the correct choice I think, since the padding is not really necessary, and this engine will accept both.

abizjak (approved):

concordium-contracts-common/src/schema.rs
soerenbf: ah, that makes sense.. couldn't figure out why it complained - thanks!

PR #101, opened 2023-07-09
Add thresholds

abizjak:
## Purpose

Move various signature thresholds from concordium-base since they are needed now in smart contracts.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-contracts-common/src/impls.rs
+++ b/concordium-contracts-common/src/impls.rs
@@ -4,6 +4,7 @@ use crate::{constants::*, schema, traits::*, types::*};
 use alloc::{boxed::Box, collections, string::String, vec::Vec};
 use collections::{BTreeMap, BTreeSet};
 use convert::TryFrom;
+use core::cmp::Ordering;
 #[cfg(not(feature = "std"))]
 use core::{
     convert, hash, marker,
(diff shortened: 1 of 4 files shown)


concordium-contracts-common/src/impls.rs
```diff
@@ -281,6 +282,95 @@ impl Deserial for AccountBalance {
...
+impl<Kind> Clone for Threshold<Kind> {
+    fn clone(&self) -> Self {
+        Self {
+            threshold: self.threshold,
+            kind:      marker::PhantomData,
+        }
+    }
+}
+
+impl<Kind> Copy for Threshold<Kind> {}
```
DOBEN, line 327: Will this be implemented?

concordium-contracts-common/src/impls.rs
```diff
@@ -281,6 +282,95 @@ impl Deserial for AccountBalance {
...
+
+impl<Kind> PartialOrd for Threshold<Kind> {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        self.threshold.partial_cmp(&other.threshold)
+    }
+}
+
+impl<Kind> Ord for Threshold<Kind> {
+    fn cmp(&self, other: &Self) -> Ordering { self.threshold.cmp(&other.threshold) }
+}
```
DOBEN, line 346: Please add empty line.

concordium-contracts-common/src/impls.rs
```diff
@@ -281,6 +282,95 @@ impl Deserial for AccountBalance {
...
+impl<Kind> PartialOrd<u8> for Threshold<Kind> {
+    fn partial_cmp(&self, other: &u8) -> Option<Ordering> { self.threshold.partial_cmp(other) }
+}
+
+impl<Kind> PartialEq<u8> for Threshold<Kind> {
+    fn eq(&self, other: &u8) -> bool { self.threshold == *other }
+}
+
+impl<Kind> Eq for Threshold<Kind> {}
```
DOBEN, line 355: Will this be implemented?

concordium-contracts-common/src/types.rs
```diff
@@ -392,6 +393,43 @@ impl AccountBalance {
...
+/// The minimum number of credentials that need to sign any transaction coming
+/// from an associated account.
+pub type AccountThreshold = Threshold<AccountKind>;
+
+#[derive(Debug)]
+pub enum SignatureKind {}
+
+/// The minimum number of signatures on a credential that need to sign any
+/// transaction coming from an associated account.
+pub type SignatureThreshold = Threshold<SignatureKind>;
```
DOBEN, line 408: If I understand correctly, then `AccountThreshold` comes from `base` (old definition) while the `SignatureThreshold` from `concordium-std`. Both types behave technically the same but we cannot unify them because different `signature types` are already defined in different places.

Maybe we can add a note here that these two things are the same.

concordium-contracts-common/src/impls.rs
```diff
@@ -281,6 +282,95 @@ impl Deserial for AccountBalance {
...
+        }
+    }
+}
+
+impl<Kind> TryFrom<u8> for Threshold<Kind> {
+    type Error = ZeroSignatureThreshold;
+
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        if value == 0 {
+            Err(ZeroSignatureThreshold)
```
DOBEN, line 308: This type has a generic `Threshold` name (meaning people could think to use it for all kinds of thresholds), but since it returns a very specific error when the value is set to 0 (referencing signatures), the `threshold` is somewhat meant for a specific use case. 

Suggestion: Would it make it clearer if we call it `SignatureThreshold` instead of `Threshold`?

I assume you called it just `Threshold` because it is difficult to pick to either call it `SignatureThreshold` or `AccountThreshold`.

DOBEN:
Please don't forget to address the `TODO: changelog/version` update.

concordium-contracts-common/src/impls.rs
abizjak: It is implemented `Eq` has no members, it is just a marker trait that says `a == a` will always yield `true`.

concordium-contracts-common/src/impls.rs
abizjak: This is a special marker trait, this implements it, there are no additional member functions to implement. I know it looks a bit odd though. https://doc.rust-lang.org/std/marker/trait.Copy.html

It's functionality is the same as `.clone`, except implementing this tells the compiler to do it implicitly.

concordium-contracts-common/src/types.rs
limemloh, line 408: I think the name needs to tie it to the credential, just like `AccountThreshold` represents the threshold per account.
```suggestion
pub type CredentialThreshold = Threshold<CredentialKind>;
```

limemloh (approved):

concordium-contracts-common/src/types.rs
limemloh: I see now that this type is also used for accounts in base, and changing it would be breaking, so nvm

abizjak: No, both come from concordium-base. I've added a comment about it though.

concordium-contracts-common/src/impls.rs
abizjak: Ok, Renamed to `NonZeroThresholdU8`.

concordium-contracts-common/src/types.rs
```diff
@@ -392,6 +393,43 @@ impl AccountBalance {
     pub fn available(&self) -> Amount { self.total - cmp::max(self.locked, self.staked) }
 }
 
+#[derive(Debug)]
+pub enum AccountKind {}
```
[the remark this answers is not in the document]
abizjak: Added

concordium-contracts-common/src/types.rs
```diff
@@ -392,6 +393,43 @@ impl AccountBalance {
...
+
+/// The minimum number of credentials that need to sign any transaction coming
+/// from an associated account.
+pub type AccountThreshold = Threshold<AccountKind>;
+
+#[derive(Debug)]
+pub enum SignatureKind {}
```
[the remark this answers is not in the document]
abizjak: Added

concordium-contracts-common/src/types.rs
```diff
@@ -392,6 +393,43 @@ impl AccountBalance {
...
+#[cfg_attr(
+    feature = "derive-serde",
+    serde(
+        bound(serialize = "Kind: Sized", deserialize = "Kind: Sized"),
+        try_from = "u8",
+        into = "u8"
+    )
+)]
+#[derive(Debug)]
+pub struct Threshold<Kind> {
```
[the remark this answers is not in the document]
abizjak: Added documentation to the type.

PR #103, opened 2023-07-26
Add signature and key types to concordium-contracts-common

DOBEN:
## Purpose

Signature and key types are not solely used by smart contracts anymore. It makes sense to move them to `concordium-contracts-common` from `concordium-std` so they are easier accessible and can be imported by e.g. backends to construct input parameters.

## Changes

Add signature and key types to `concordium-contracts-common`

### DIFF
--- a/concordium-contracts-common/src/impls.rs
+++ b/concordium-contracts-common/src/impls.rs
@@ -372,6 +372,48 @@ impl<Kind> NonZeroThresholdU8<Kind> {
     };
 }
 
+impl Serial for PublicKeyEd25519 {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> { self.0.serial(out) }
+}
+
+impl Deserial for PublicKeyEd25519 {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        Ok(PublicKeyEd25519(Deserial::deserial(source)?))
+    }
+}
+
+impl schema::SchemaType for PublicKeyEd25519 {
+    fn get_type() -> crate::schema::Type { schema::Type::ByteArray(32) }
+}
+
+impl Serial for PublicKeyEcdsaSecp256k1 {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> { self.0.serial(out) }
+}
+
+impl Deserial for PublicKeyEcdsaSecp256k1 {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        Ok(PublicKeyEcdsaSecp256k1(Deserial::deserial(source)?))
+    }
+}
+
+impl schema::SchemaType for PublicKeyEcdsaSecp256k1 {
+    fn get_type() -> crate::schema::Type { schema::Type::ByteArray(33) }
+}
+
+impl Serial for SignatureEd25519 {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> { self.0.serial(out) }
+}
+
+impl Deserial for SignatureEd25519 {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        Ok(SignatureEd25519(Deserial::deserial(source)?))
+    }
+}
+
+impl schema::SchemaType for SignatureEd25519 {
+    fn get_type() -> crate::schema::Type { schema::Type::ByteArray(64) }
+}
+
 impl Serial for ExchangeRate {
     fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> {
         out.write_u64(self.numerator())?;
--- a/concordium-contracts-common/src/impls.rs
+++ b/concordium-contracts-common/src/impls.rs
@@ -372,56 +372,16 @@ impl<Kind> NonZeroThresholdU8<Kind> {
     };
 }
 
-impl Serial for PublicKeyEd25519 {
-    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> { self.0.serial(out) }
-}
-
-impl Deserial for PublicKeyEd25519 {
-    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
-        Ok(PublicKeyEd25519(Deserial::deserial(source)?))
-    }
-}
-
 impl schema::SchemaType for PublicKeyEd25519 {
     fn get_type() -> crate::schema::Type { schema::Type::ByteArray(32) }
-}
-
-impl Serial for PublicKeyEcdsaSecp256k1 {
-    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> { self.0.serial(out) }
-}
-
-impl Deserial for PublicKeyEcdsaSecp256k1 {
-    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
-        Ok(PublicKeyEcdsaSecp256k1(Deserial::deserial(source)?))
-    }
 }
 
 impl schema::SchemaType for PublicKeyEcdsaSecp256k1 {
     fn get_type() -> crate::schema::Type { schema::Type::ByteArray(33) }
 }
 
-impl Serial for SignatureEd25519 {
-    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> { self.0.serial(out) }
-}
-
-impl Deserial for SignatureEd25519 {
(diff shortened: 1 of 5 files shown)


concordium-contracts-common/src/impls.rs
```diff
@@ -372,6 +372,62 @@ impl<Kind> NonZeroThresholdU8<Kind> {
...
 
+impl Serial for PublicKeyEd25519 {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> { self.0.serial(out) }
+}
+
+impl Deserial for PublicKeyEd25519 {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        Ok(PublicKeyEd25519(Deserial::deserial(source)?))
+    }
+}
```
abizjak, line 383: We could just derive this

concordium-contracts-common/src/impls.rs
```diff
@@ -372,6 +372,62 @@ impl<Kind> NonZeroThresholdU8<Kind> {
...
+
+impl schema::SchemaType for PublicKeyEd25519 {
+    fn get_type() -> crate::schema::Type { schema::Type::ByteArray(32) }
+}
```
abizjak, line 387: Not this though.

concordium-contracts-common/src/impls.rs
```diff
@@ -372,6 +372,62 @@ impl<Kind> NonZeroThresholdU8<Kind> {
...
+
+impl Serial for PublicKeyEcdsaSecp256k1 {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> { self.0.serial(out) }
+}
+
+impl Deserial for PublicKeyEcdsaSecp256k1 {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        Ok(PublicKeyEcdsaSecp256k1(Deserial::deserial(source)?))
+    }
+}
```
abizjak, line 397: Derive this.

concordium-contracts-common/src/impls.rs
```diff
@@ -372,6 +372,62 @@ impl<Kind> NonZeroThresholdU8<Kind> {
...
+
+impl Serial for SignatureEd25519 {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> { self.0.serial(out) }
+}
+
+impl Deserial for SignatureEd25519 {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        Ok(SignatureEd25519(Deserial::deserial(source)?))
+    }
+}
```
abizjak, line 411: Derive this.

concordium-contracts-common/src/impls.rs
```diff
@@ -372,6 +372,62 @@ impl<Kind> NonZeroThresholdU8<Kind> {
...
+
+impl Serial for SignatureEcdsaSecp256k1 {
+    fn serial<W: Write>(&self, out: &mut W) -> Result<(), W::Err> { self.0.serial(out) }
+}
+
+impl Deserial for SignatureEcdsaSecp256k1 {
+    fn deserial<R: Read>(source: &mut R) -> ParseResult<Self> {
+        Ok(SignatureEcdsaSecp256k1(Deserial::deserial(source)?))
+    }
+}
```
abizjak, line 425: And this.

concordium-contracts-common/src/types.rs
```diff
@@ -448,6 +450,211 @@ pub struct NonZeroThresholdU8<Kind> {
...
+/// single scheme is supported, `ed25519`.
+pub enum PublicKey {
+    Ed25519(PublicKeyEd25519),
+}
+
+#[cfg(feature = "std")]
+#[derive(crate::Serialize, Debug, crate::SchemaType, PartialEq, Eq)]
+pub struct CredentialPublicKeys {
+    #[concordium(size_length = 1)]
+    pub keys:      crate::collections::BTreeMap<KeyIndex, PublicKey>,
```
abizjak, line 608:
```suggestion
    pub keys:      BTreeMap<KeyIndex, PublicKey>,
```

It's already imported in this module (in both std and no-std variants).

concordium-contracts-common/src/lib.rs
```diff
@@ -83,6 +83,16 @@
 #[cfg(not(feature = "std"))]
 extern crate alloc;
 
+/// Re-export.
+#[cfg(feature = "std")]
+pub use std::{string::String, vec::Vec};
```
abizjak, line 88: I don't think we should be re-exporting things in contracts-common.

We do it in concordium-std to provide an easy experience for users of that library, but contracts-common is different.

If you look at the comments I have below, this is not necessary.

concordium-contracts-common/src/lib.rs
```diff
@@ -83,6 +83,16 @@
...
+
+/// Re-export.
+pub mod collections {
+    #[cfg(feature = "std")]
+    pub use std::collections::BTreeMap;
+}
+
```
abizjak, line 95: I don't think this is necessary.

concordium-contracts-common/src/lib.rs
```diff
@@ -97,7 +107,7 @@ mod types;
 #[cfg(feature = "smart-contract")]
 pub use concordium_contracts_common_derive::*;
 #[cfg(not(feature = "smart-contract"))]
-pub use concordium_contracts_common_derive::{Deserial, Serial, Serialize};
+pub use concordium_contracts_common_derive::{Deserial, SchemaType, Serial, Serialize};
```
abizjak, line 110: Why this? Typicallly schematype should only be needed in smart contracts, at least it should only be derived there.

concordium-contracts-common/src/types.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(feature = "std")]
+use crate as concordium_std;
```
abizjak, line 2: Why do you need this only when `std` feature is enabled? I would have assumed you need it always.

concordium-contracts-common/src/types.rs
```diff
@@ -448,6 +450,211 @@ pub struct NonZeroThresholdU8<Kind> {
...
+
+#[cfg(feature = "std")]
+#[derive(crate::Serialize, Debug, crate::SchemaType, PartialEq, Eq)]
+/// Public keys of an account, together with the thresholds.
+/// This type is deliberately made opaque, but it has serialization instances
+/// since inside smart contracts there is no need to inspect the values other
+/// than to pass them to verification functions.
+pub struct AccountPublicKeys {
+    #[concordium(size_length = 1)]
+    pub keys:      crate::collections::BTreeMap<CredentialIndex, CredentialPublicKeys>,
```
abizjak, line 620:
```suggestion
    pub keys:      BTreeMap<CredentialIndex, CredentialPublicKeys>,
```

concordium-contracts-common/src/types.rs
DOBEN: Unfortunately, the `BTreeMap` is not imported yet, only when `std` feature is explicitly enabled

e.g. this was the import statement before
```
#[cfg(all(feature = "std", feature = "concordium-quickcheck"))]
use std::collections::BTreeMap;
```

Most of the added types now only compile with the `std` feature enabled, e.g.
```
#[cfg(feature = "std"))]
use std::collections::BTreeMap;
```

These new key and signature types are now compiled only when `std` feature is explicitly enabled because we need the `BTreeMap` as well as since we derive `SchemaType` we need `std::{string::String, vec::Vec}` in the crate root.

I acknowledge that this is not all we want (we would like to have these types available also in `no-std` build) but at the same time I want the `SchemaType` to be derived for these types.

concordium-contracts-common/src/lib.rs
DOBEN: I don't re-export them anymore but I still import them here since `SchemaType` needs it.

concordium-contracts-common/src/lib.rs
DOBEN: Some of the `PublicKey, CredentialPublicKeys, ...` derive it but I moved the import into the `types.rs` file since it is used there.

concordium-contracts-common/src/types.rs
DOBEN: removed it

concordium-contracts-common/src/types.rs
abizjak: I pushed changes fixing this now.

abizjak (approved):

